PR #203, opened 2018-10-17
Update README.md

Poussinou:
Description:
This small PR adds a button to get your app on F-Droid and another one for Google Play

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,12 @@ This is as good as it sounds. It's what we've been waiting for.
 
 ## Installation
 
-The client is available on google play. Following the link below you will be redirected to the store page.
-[Secure Quick Reliable Login](https://play.google.com/store/apps/details?id=org.ea.sqrl)
+[<img src="https://f-droid.org/badge/get-it-on.png"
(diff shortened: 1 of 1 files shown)


kalaspuffar:
Looks like a good pull request. I would put them in another sort order.

And the link to unknown sources should be under a subheading like
"We recommend that you download apps from Google Play. You can also get them from other sources."

Does that sound acceptable?

Poussinou:
Yep, sounds like something acceptable 👍 Would probably be better if you do it yourself, I understand that you prefer Google Play as the way to get the app.

kalaspuffar:
Hi @Poussinou 

I've updated the readme with the link in question.

Just to give my 1 cent, I do not prefer Google Play. Prefer is a strong word. As the f-droid site states "All the packages should be installed at your own risk". I feel safer to recommend a platform that Google is actively scanning for malicious applications.

F-Droid seems like a great project that I probably will support in the future, never installed anything from it but I'm glad there is a platform where free software can be found.

Best regards
Daniel

PR #207, opened 2018-11-15
compress background.png (webp and jpg)

sesam:
Tested on Samsung S5.
Ready to be merged.

Learnt for next time :) - if we take an original jpg and crop and re-optimize in one step, we could get better quality per byte ratio. Maybe worth doing if a tablet app design is implemented in the future.


kalaspuffar:
Well, cropping could be something to look into. Then again I want to be able to support different orientations with just one background image so it needs to be a bit larger. I've tried the app on my tablets and so far the background doesn't show that many artifacts. Will try with this optimized version. Have you tried webp optimization? Maybe it will be as good as lossless compression. 

Then again does all Android versions support it?

Sorry, didn't see this PR. Will test and merge it ASAP.

kalaspuffar:
The http://squoosh.app could be a good testing ground.

sesam:
ping: tested and mergable?

kalaspuffar:
I tried webp and vector drawable today.

Works fine on a new device.
Want to test on a old device as well.
90% reduction without loss in detail.

kalaspuffar:
Been a lot of work and other required engagements.

I wanted to test webp to see if it is a viable option for older devices. If I get time this evening I'll try it on older device.

If not a viable option or if this image has a better compression to artifact ratio I'll merge this.

Issue #244, opened 2019-02-15
Cannot paste password


janneoksanen:
After I scan a QR code the app prompts for my password. My password is super complex so there is no way I'm typing it in every time I need to log in to a service. I can copy the password from my password manager but I cannot paste it in the password field.

kalaspuffar:
Shouldn't the password manager be able to fill in the password?

PR #254, opened 2019-03-01
Long click listener to paste password

sengsational:
Issue #244 

Description:
Listener on password field that pastes clipboard contents

Changes:
Added long click listener that pastes the contents of the clipboard if the user does a long click on the password field.  A short toast confirms the paste if successful.  If the clipboard is empty or contains something short, it doesn't do anything (except some logging). 

Notes:
The toast text probably should be put into `strings.xml`, but I didn't know how to get it into all languages (figured there might be automation to manage that).

This might not be the "perfect" way to implement this since instead of pasting on a long click, it could interact with the user before pasting.  

It's possible that they've already got contents in the field.  If so, it's replaced with the clipboard contents.  

The clipboard is _not_ cleared after the paste occurs because ClipboardManager.clearPrimaryClip() throws a NoSuchMethodError, and I didn't see an alternative.


sengsational:
This pull request is closed because a later one (#257) covers what's in this request.

Issue #220, opened 2018-12-09
feat: make current identity visible on main login screen


IvarSnaaijer:
Currently we can not see the current identity used on the main screen. 

If I recall correctly this becomes apparent later but it still would be usefull

sesam:
While the idea with SQRL is that you only ever need one identity, at least while testing it would be practical to see what identity is being used.

The QR scan view currently shows one text "Scan QRCode for site" and maybe that is a quick and good enough place to add in a (name of identity) at the end.

Another idea is to put the chosen identity at the top of the front page where it currently says "SQRL - Main". I don't know Android design guidelines, but probably changing that header text to `(identityname) - (app name)` could be ok.

IvarSnaaijer:
I'm still conflicted. Possibly this does need to go to a three stripes menu in the top left corner (like Google does), in my current proposal it is rather prominent. You could make it clear to users that creating a new identity should not be necessary in the ´create´ window (by giving use cases), and only show the selection when there is actually something to choose.

PR #256, opened 2019-03-01
WIP: Show identity name in the ActionBar of SimplifiedActivity

janneoksanen:
Issue #220 
feat: make current identity visible on main login screen

Description:
Shows currently used identity name on the ActionBar of SimplifiedActivity

screenshot_20190301-151542


kalaspuffar:
Hi @janneoksanen 

Great work, this seems like a good quick fix for the issue. Not super clear that the information is found in the app bar. Don't know if I've seen this solution in any other app. I was thinking like a quick fix to add a panel to the icon with some text. Like in this wireframe.

https://s3.amazonaws.com/assets.mockflow.com/app/wireframepro/company/Cc09d9f00343be5aa571cc840f3472f46/projects/Mf421dde088ed6d581c66618009e730f21551597018478/pages/e80320270a0c4609a49206bd7ebe9c09/image/e80320270a0c4609a49206bd7ebe9c09.png

I would like someone to create a good design for these pages though so we don't do too many design changes that require new text strings that translators need to translate.

What do you think, shall I add your merge request in the meantime even though it's not the final solution?

Best regards
Daniel

janneoksanen:
Hi Daniel, 

I'll do it according to the wireframe so we don't have to go back and fix the fix.

Janne

janneoksanen:
How about something like this?

screenshot_20190308-110937

kalaspuffar:
Hi Janne

Great work, not sure if we want to introduce another blue color. Maybe we should keep it simple with black on white instead.

Not looked at the code yet, but I will as soon as time permits.

Best regards
Daniel

kalaspuffar:

PR #257, opened 2019-03-02
WIP: Add password paste in 2 places

sengsational:
Issue #244

Description:
Listener on *two* password fields that pastes clipboard contents

Changes:
Added long click listeners that pastes the contents of the clipboard if the user does a long click on the password fields. A short toast confirms the paste if successful. If the clipboard is empty or contains something short, it doesn't do anything (except some logging).

Notes:
The toast text probably should be put into strings.xml, but I didn't know how to get it into all languages (figured there might be automation to manage that).

This might not be the "perfect" way to implement this since instead of pasting on a long click, it could interact with the user before pasting.

It's possible that they've already got contents in the field. If so, it's replaced with the clipboard contents.

The clipboard is not cleared after the paste occurs because ClipboardManager.clearPrimaryClip() throws a NoSuchMethodError, and I didn't see an alternative.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/SettingsActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/SettingsActivity.java
@@ -1,10 +1,13 @@
 package org.ea.sqrl.activites;
 
+import android.content.ClipData;
+import android.content.ClipboardManager;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.os.Handler;
 import android.support.constraint.ConstraintLayout;
 import android.os.Bundle;
+import android.util.Log;
 import android.view.Gravity;
 import android.view.LayoutInflater;
 import android.view.View;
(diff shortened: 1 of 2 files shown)


kalaspuffar:
Hi @sengsational 

I wonder if we should implement this feature, still a discussion in the Forum. https://sqrl.grc.com/threads/pasting-the-password.399/

Steve's opinion:
"Yes. I believe that allowing cutting / copying / pasting the password or Rescue Code should be strongly discouraged. It's trivially easy for apps to monitor and capture the system clipboard. It's convenient as hell, and also insecure as hell."

Whatever we do we need to consider if there is a way to do this in the platform without implementing the feature ourselves. I know you've already done it and it seems good at first glace.

Might be a few more places we should add it and you are right there is an "automated" way for the strings. All strings are exported from POEditor so I add the term there and then export the translated strings and prepare them for Android.

Great work though and if we want this feature and there isn't another way I'll be glad to merge this feature.

Best regards
Daniel

sengsational:
Thanks for the update.  I knew that this was going to be a trade-off between security and convenience.  I'll do some research to better understand the exposure in the Android environment.

kalaspuffar:

sengsational:
Based on Steve's assessment, I think this pull request can be deleted / removed / ignored, or otherwise transitioned into where ever pull requests go to die.

Issue #252, opened 2019-02-21
Improve login experiance


kalaspuffar:
Reported by @sengsational

On the Android app, I clicked "Advanced Functions > Use Identity" (probably needs to be moved to the main page at some point), and it was obvious I needed to scan the QR code on this site's registration page. The app flipped back out of camera mode. Then it's asking for a "password". Here I found myself wondering "does this app think I already have a password for sqrl.grc.com? And I'm thinking "whoa, I'm not giving my SQRL master password to this hinky forum site, who knows who this Gibson character is, anyway" ;) Seriously, the point I'm trying to make is that instead of "password", it should probably be labeled "SQRL Master Password" and have an info thing that pops up and says "your master password is used to decrypt your identity...it won't leave your phone". Of course I realize that this UI / usability stuff is icing on the cake, but since I have "new eyes" today, I figured I'd write my "new eyes" reaction to today's version of the application. Certainly no malice intended.

kalaspuffar:
Change label and perhaps add text to login button.

IvarSnaaijer:
I opt for a label "Login with:" and the currently selected handle.

kalaspuffar:
Hi @IvarSnaaijer

Sounds great.

I've added the term in POEditor but ended my dev time for today. 

Added alternative identity as a first draft. :)

Best regards
Daniel

kalaspuffar:
![](https://s3.amazonaws.com/assets.mockflow.com/app/wireframepro/company/Cc09d9f00343be5aa571cc840f3472f46/projects/Mf421dde088ed6d581c66618009e730f21551597018478/pages/e80320270a0c4609a49206bd7ebe9c09/image/e80320270a0c4609a49206bd7ebe9c09.png)

IvarSnaaijer:
Yes, especially the box will inform the user in a pleasant way that this is actually a big button. The new label (Tools/Setup) also is an improvement. I hope to find those soon in PO Edit.

sesam:
A suggestion: buttons usually have the text under the picture, so if we want people to understand this as a button and not a modal-looking dialog window, it probably should look more like https://www.google.com/search?q=android+icon+button+with+text

sengsational:
When I wrote:

>  I clicked "Advanced Functions > Use Identity" (probably needs to be moved to the main page at some point)

That was because I didn't understand that the big SQRL logo on the main page was actually a button, hehe!

IvarSnaaijer:
[...]

PR #259, opened 2019-03-09
Add scan qr icon and associated button

sengsational:
Issue #252  Improve login experience
Description:
Created an icon and associated button on mail activity.'

Changes:
Added `<Button...` which uses a new vector graphic icon of a magnifying glass aiming at a simplified QR code.
button with icon

Notes:
This change was as much for training me how to use the source control that's integrated with Android Studio as anything.  I have no doubt that there's some "real" UI people out there that will give us guidelines, but we could use this in the mean time.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/SimplifiedActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/SimplifiedActivity.java
@@ -3,11 +3,13 @@ package org.ea.sqrl.activites;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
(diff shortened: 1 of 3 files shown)


sengsational:
Changed from `<Button...` to `AppCompatButton...` and also had to remove `android:drawableLeft...` since that causes problems on older versions of the OS.

kalaspuffar:

sengsational:
This is not a part of the comprehensive UI update, but might make it more clear for new users what to do on the simplified activity home screen.

sengsational:
Fixed a small syntax error in the layout.

PR #275, opened 2019-03-31
Move quickpass to preferences.

kalaspuffar:
Issue #274

Description:
In order to keep the quickpass when the application is removed from memory, I will save the quickpass data in preferences.

Changes:
Added code in SQRLStorage to save quickpass data to preferences as a hex string and when it requires use fetch the string and decode it. This requires that the storage class has a context so I reuse the app contexts when the storage is created.
The same context is used to remove quickpass and show eventual popup messages. This might not work depending if the context needs to be live in order to allow this.


kalaspuffar:
Hi @sengsational

Could you test and review the code?

Best regards
Daniel

sengsational:
I will give it a try tomorrow.

sengsational:
I found the git "fetch" command, which made the quickpass_preferences branch show in my AndroidStudio, so I though "I must do a quick test tonight".  The test was successful!

I performed the operations to my OS not to save background processes (see: https://sqrl.grc.com/threads/android-v0-13-5.406/post-3364)

I aimed the app at the sqrl.grc.com QR code, used full password, successfully logged on (same/expected).  Then I switched away from the application and swiped it off the active apps list.  The small icon remained in the upper left (same/expected).  I logged off the forums site.  I opened the app again, it took a bit longer to get going (as expected since it had to restart everything).  Aimed at the QR code requested quick pass (as expected and an _improvement over the previous version_), after quick pass was entered, successfully logged on.

I haven't looked at the code, but this one test indicates that quickpass will now survive even if the OS forces the app to rebuild itself.

kalaspuffar:
Hi @sengsational 

Do you want to look at the code before I merge?

Best regards
Daniel

sengsational:
Wow, lot's of changes, but only because of changing how the context is passed. 

I see the sharedPreferences addition in decryptIdentityKeyQuickPass().  Looks nice.

I would feel better if we did not have the instance variable `SQRLStorage.context` at all.  We only use it inside the quick pass related methods and to get to `PreferenceManager` and `NotificationManager`, and the processes that call those methods have their own good context.  I see that if the OS removes our instance variable `instance` we will populate `context` when the singleton is recreated, so it should be fine.  I'm just not very comfortable about how Android does these rebuilding things, and like to avoid any possibility that things could be null when we are depending on them.

In summary, because I tested with background processes "off", and it worked, I'm comfortable with this, despite the instance variable.

kalaspuffar:
Hi @sengsational 

Well, I don't like this solution either. But passing it in every call that we use quickpass is not a good solution either. Maybe the instance function should refresh the context every time?

I've realized that I need to add the biometrics to preferences as well to make the fingerprint survive between app cycles. Another way to handle these things is to save them using StateBundle and saving them on every pause and resume and get the state in each activity and store that state in preference. Might be a bit more work though.

Best regards
Daniel

PR #276, opened 2019-04-01
doLogin code cleanup

sengsational:
Description:
Cleaned up duplicate code in login process.  Added submit password directly from keyboard.

Changes:
Replaced common code, 2 from UrlLoginActivity, 2 from LoginBaseActivity, and 1 from AlternativeLoginActivity with `doLogin()`.  The intent was to preserve all functionality (not change anything).

Added listener so that the user doesn't have to first dismiss the keyboard then hit a button when logging on.  This tiny feature was where I started, but I didn't want to duplicate the code another time!

Discussion:
I realize that this is kind of a scary big set of changes, but I pulled all five instances of this duplicate code out and did a "diff".  I think I've got all the functionality intact.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
@@ -13,6 +13,7 @@ import android.text.Editable;
 import android.text.TextWatcher;
 import android.util.Log;
 import android.view.Gravity;
+import android.view.inputmethod.EditorInfo;
 import android.widget.EditText;
 import android.widget.TextView;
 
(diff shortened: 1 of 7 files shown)


sengsational:
I wasn't completely sure why `hideLoginPopup()` and `showLoginPopup()` were managed with some logons and not others, but it looked like it correlated with the CPS type.  

There are cases where the `setHint()` call doesn't change the state, but that doesn't hurt anything.

Hopefully I've got this right.  I'll understand if you don't have time to analyze this!!

kalaspuffar:
Could you resolve the conflicts in this pull request before we look at it closer?

sengsational:
This pull request has been replace by #286

PR #277, opened 2019-04-01
Animate simple button.

kalaspuffar:
Issue #252.

Description:
Animate the login button until the user clicks on it the first time to indicate actionable item.

Changes:
Added animation.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/SimplifiedActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/SimplifiedActivity.java
@@ -7,8 +7,11 @@ import android.hardware.biometrics.BiometricPrompt;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.CancellationSignal;
+import android.preference.PreferenceManager;
 import android.support.design.widget.Snackbar;
 import android.util.Log;
+import android.view.animation.Animation;
+import android.view.animation.AnimationUtils;
 import android.widget.Button;
 import android.widget.ImageButton;
 import android.widget.TextView;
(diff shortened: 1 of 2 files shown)


kalaspuffar:
Hi @sengsational 

Can you look at this pull request and see if it works and if you like the presentation.

Best regards
Daniel

sengsational:
Tested on my physical device API 27 and it shows the button pulsating, as you intended.

I'm afraid I'm "too close" to the application to know whether that animation would help a new user identify the big blue squirrel as a "button".  I'm fairly sure it would, but it would also annoy people :)

I think one problem is that the graphic does not have the "depth" and "shadow" of a button.  Such as this example: https://rfclipart.com/image/big/67-6c-6a/blank-green-web-buttons-with-different-shadows-Download-Royalty-free-Vector-File-EPS-100337.jpg

I still think something like 
magqr
more precisely tells the user what's going to happen when they press it.  That vector image is part of the scan_qr_icon_button branch.

kalaspuffar:
Hi @sengsational 

Well then again it should be annoying. :)

It only animates if you haven't clicked it before. Then we will stop doing that animation so I think it's fine.  like your icon but I've not seen how it would look on without background or if we could design it into the middle with text beneath. 

The scan_qr_icon_button is also good but I don't want two buttons doing the same thing on screen. The whole point with that screen is that it should be simplified. I'm thinking to move the Advanced options button to the menu instead. That would clear up some of the UI leaving the only option on screen which would entice people to use it.

Maybe we can do a mix of your branch and mine? Maybe skip the animation, big button front center.

Thoughts?

Best regards
Daniel

sengsational:
I like what you've done.

Issue #280, opened 2019-04-01
Back after identity creation


kalaspuffar:
When you create a identity for the first time and are done at the simplified view. Then when you press back in on your phone you go back to the introduction screen.

Expected: should close application

PR #281, opened 2019-04-01
Fix bug where the creating process goes to the wrong activity.

kalaspuffar:
Issue #280 .

Description:
When you created your first identity and press back you should close app. 
Today you can go back to the introduction screen. Should not be available when identity is present.

Changes:
Close current activity and all other running activities and then fire up the simplified view so we end up in the same state as when you open the app with identities loaded.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/create/NewIdentityDoneActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/create/NewIdentityDoneActivity.java
@@ -9,6 +9,7 @@ import android.widget.Button;
 import android.widget.TextView;
 
 import org.ea.sqrl.R;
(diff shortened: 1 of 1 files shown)


kalaspuffar:
Hi @sengsational 

Can you test and review the code please.

Best regards
Daniel

sengsational:
I see that the task is taking the default `launchMode`, but I don't know enough about the differences between `finishAffinity()` and `finishRemoveTask()` to add any value to the discussion.

Although I can occasionally get the emulator working, today it doesn't want to work properly, so couldn't test on API 16.   On API 27 (my physical phone) my testing was successful (after new identity > done > simplified login screen) as expected.

PR #286, opened 2019-04-08
Consolidate login

sengsational:
Description:
Replace near-duplicate code in 5 places with a method to manage all.

Changes:
Wrote a `doLogin()` method that replicated the login code in the 5 places.  Passed in booleans to manage the differences.


sengsational:
This pull request should only include 9198098.

I can't figure out how to get rid of 5ebcc50 and get rid of e164000.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
```diff
@@ -103,45 +103,7 @@ public void beforeTextChanged(CharSequence password, int start, int count, int a
...
-                            handler.post(() -> {
-                                txtLoginPassword.setHint(R.string.login_identity_password);
-                                txtLoginPassword.setText("");
-                                hideProgressPopup();
-                            });
-                            storage.clear();
-                            storage.clearQuickPass();
-                            return;
-                        }
-
```
sengsational: In this code, if `decryptionOk`, there is no `showClearNotification()`, which the reason for the boolean.  Instance 1 of 2.

app/src/main/java/org/ea/sqrl/activites/base/LoginBaseActivity.java
```diff
@@ -240,47 +201,58 @@ public void afterTextChanged(Editable s) {
...
+                    txtLoginPassword.setHint(R.string.login_identity_password);
+                    txtLoginPassword.setText("");
+                    hideProgressPopup();
+                    showLoginPopup();
+                });
+                storage.clear();
+                storage.clearQuickPass();
+                return;
+            }
+            if(!usedCps) showClearNotification();
```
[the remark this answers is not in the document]
sengsational: I made a note in the code where `showClearNotification()` wasn't being called.

This would change the behavior compared to existing code, but I can see how it needs to be removed so it will always be called.  I'll work on it.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
kalaspuffar: One of the unfortunate inconsistencies :(

If you succeed in decrypting the identity we will set the quickpass and therefore we need to show the clear quickpass notification in the notification bar.

app/src/main/java/org/ea/sqrl/activites/base/LoginBaseActivity.java
```diff
@@ -240,47 +209,58 @@ public void afterTextChanged(Editable s) {
...
+
+        new Thread(() -> {
+            boolean decryptionOk = storage.decryptIdentityKey(txtLoginPassword.getText().toString(), entropyHarvester, usedQuickpass);
+            if(!decryptionOk) {
+                showErrorMessage(R.string.decrypt_identity_fail);
+                handler.post(() -> {
+                    txtLoginPassword.setHint(R.string.login_identity_password);
+                    txtLoginPassword.setText("");
+                    hideProgressPopup();
+                    showLoginPopup();
```
kalaspuffar: showLoginPopup is only when we don't login to a URL click ( !usedCPS)

sengsational: Added a conditional so that the popup won't show at the wrong time.

Issue #278, opened 2019-04-01
strength indicator


kalaspuffar:
Add password strength indicator to show password strength. Don't add any requirement for password length just recommendation.

silversword411:
Not sure how to implement, but these look like 3 different possibilities:
https://github.com/yesterselga/password-strength-checker-android
https://blog.davidmedenjak.com/android/2017/01/14/password-strength.html
http://crazygeeksblog.com/2017/10/password-strength-validation-checker-in-android-with-example/

kalaspuffar:
Hi @silversword411 

Great input. Seems like the last link is pretty much what I was thinking of. Having a progress bar showing some strength indication maybe with some colors and perhaps some descriptive text.

And then make some quantitive calculation on the strength. Not keen on adding a library to solve this "trivial" problem. Maybe we can take inspiration from GRC password strength check.

https://www.grc.com/haystack.htm

Best regards
Daniel

alexhauser:
Hi,

I am working on this and a PR should come in shortly.

However, this is not as "trivial" as it looks at first sight. Measuring password strength only by password length and use of character types (upper/lower/digit/symbol) like Steve's password haystack would fail to detect a whole class of really bad passwords, like commonly used passwords, common keyboard patterns, guessable passwords etc. Steve even notes in big letters that password haystack is NOT a password strength meter.

Now, the most robust thing would probably be to include a library like https://github.com/nulab/zxcvbn4j, but it is large and supposedly also slow, so I guess it would probably be overkill.

That's why I am working on a hybrid solution, using a simple algorithm based on password length and character classes, but additionally checking against a list of the 10.000 most common passwords from https://github.com/danielmiessler/SecLists.

What do you guys think?

Here are some preliminary screenshots:
![01](https://user-images.githubusercontent.com/4005543/55961901-23e7d800-5c70-11e9-8b88-f00b9de78058.JPG)
![02](https://user-images.githubusercontent.com/4005543/55961902-24806e80-5c70-11e9-9252-0091e30d0519.JPG)
![03](https://user-images.githubusercontent.com/4005543/55961903-24806e80-5c70-11e9-8ec4-19c8aced3178.JPG)
![04](https://user-images.githubusercontent.com/4005543/55961904-24806e80-5c70-11e9-925f-60b356d4d37f.JPG)

jroedel:
Sounds like a great idea to me!

[...]

PR #289, opened 2019-04-12
Feature: Password strength meter

alexhauser:
Issue #278

Calculate password strength and display a password strength meter. Also, compare the password against a list of commonly used, unsafe passwords.

Main components are 

- the _password_strength_meter.xml_ layout which can be included into any activity/fragment containing a password field. 

- the _PasswordStrengthMeter_ class which ties together the password TextEdit field and the strength meter layout.

I tried to keep this as self-contained and reusable as possible, so that we can add it in any number of activities/fragments throughout the app. For the time being, I only added the feature to the _SaveIdentityActivity_.

The "algorithm" for calculating the password strength is currently very naive and cannot be compared to solutions like https://github.com/nulab/zxcvbn4j. Any discussion on possible improvement in this area is highly appreciated. The same goes for every other aspect of the code as well of course!

Looking forward to hearing what you guys think! 

pw_strength_screenshot


sengsational:
Should the app simply disallow passwords that are not at least 6 characters?

I was reading the 10K password list (quite entertaining), but many of them are 6 or fewer characters.  We could make the list much smaller by implementing the 6 or more rule.

alexhauser:
Thanks for your input, @sengsational .

As for the app disallowing passwords under 6 characters: 
I have followed @kalaspuffar 's specification in issue #278 not to add any hard requirement. Me personally, as much as I hate weak passwords, I also don't think that we should force such a rigid rule on the users. If we're telling them that their password is poor but they still want to use it, the shame is on them I guess. But I could imagine only having a "password too short" notification instead of the "password is on list" notification for all 6-or-less-character passwords. This way, we could get rid of the short passwords on the list but still avoid a hard requirement.

The common password list:
The list comes from https://github.com/danielmiessler/SecLists, and there are MANY other options for better and more extensive lists there. But I had to make a choice, taking into account file size (not bloating the app), in-memory size, computation time for loading the list as well as for comparison with the actual password. (The passwords are loaded into a HashSet<String> at load time so that they can be quickly searched for matches every time the password field changes)

There are also many options on how to optimize the list. Using lowerCase() is a great idea, and I think I should implement it, because it's cheap. I could also build in-memory mutations of any given password on the list, like for example common substitutions of characters with numbers or symbols ( A -> 4, a -> @, S -> $ ...). This would only increse the memory footprint without bloating the apk.

Thanks again!

alexhauser:
Implemented @sengsational 's _toLowerCase()_ suggestion, waiting for further input on the 6-character topic.

kalaspuffar:

My thought is what the list actually adds to the security. Yes I know it's extra bad to use common passwords but we already mark passwords as bad if we don't use special characters, upper, lower and numbers. So the case where you use a common password would only say poor anyway.

One objection to the list is the size of the APK but I think zip will handle that reasonably well so that is no matter. But another thing that I'm thinking about is lude or rude language. I know that people choose passwords with bad words just to spite other people or to make it "more secure". 

But as a maintainer to an open source project, do we really want to have a list of potentially bad words that people could take offense to that don't really add to the security to the passwords?

If I choose a password that is long enough and with common words it's still secure.

For example

Th1sIsMY5uperSecureP@ssword!!!Take4St46AtItify0ulik3

There are many common words in there but I would say that it's still pretty secure. So do the password list add enough to the meter?

Best regards
Daniel

alexhauser:
Thanks for sharing your thoughts, @kalaspuffar .

I agree that in its current, stripped-down form, the password list is only of limited value, because as you say, most of its entries are flagged by the algorithm anyway. But in general, I think a common passwords list is important, because I'm sure there are still enough passwords on the larger lists that would rate "medium" or "good" in the algorithm but are still crackable within seconds.

Also, we do not punish having common words concatenated together. So your example above would rate perfectly fine. Only exact matches against the list (after toLowerCase() ) are regarded.

I suggest I give this another try. I'll hand-craft the list to contain only entries that would not be flagged by our algorithm anyway and that add real value to the meter, and use zip on it.

If you still aren't convinced of the list afterwards, I'll happily throw it out. ;-)
Can you please let me know what you regard as the size limit for the (zipped) list?

alexhauser:
Ok, I've reworked the whole thing:

- There is a "short password" warning now for all passwords with a length of only 1-7 characters
- The common password list now contains only entries that would have been rated medium or good by the algorithm and that are longer as the above soft min size
- The common password list is now zipped
- The common password list is now larger (77k entries instead of 10k)

This now catches tons of stuff like _password123_, _12345678910_, _qwertyqwerty_, _harrypotter_...

The zip now "wheighs" 524k, loading it to memory adds ~8mb to the memory footprint of the activity, if the android studio profiler is to be trusted. Please let me know if that is acceptable to you folks and also what you think of the whole thing now.

Thanks and have a great start into the weeek, 
Alex

kalaspuffar:
Hi @alexhauser 

The zip was not an issue per say. The APK is a zip archive so it would compress nicely without us doing anything. It was just some rambling about the size not really mattering.

I've looked through the password list and we still have a lot of offensive words in it.

I believe we can skip the list. It's usually a good feature for a site so multiple users don't choose the same password and make it easier to get multiple users accounts is database goes missing.

On a phone I think it's overkill and also when we have a limit for 5 seconds per try you need a lot of time just to test the full list.

So over all I think the strength with some metrics should do fine. We will do the same as the reference client and the user might feel more secure.

It's all about image, like when you fly and goes through security :)

Best regards
Daniel

alexhauser:
Hi @kalaspuffar , 

thanks for your feedback. While I don't quite agree with the "it's all about image" notion, I agree that the imposed rate limit on decryption attempts (if it holds true even when performed offline/off-device) is a strong argument on your side. :)

We should just all be clear about the fact the password strength meter then isn't really measuring password strength, but only a specific subset thereof.

I'll remove the common password list with my next commit, which could take a day or two since I'm travelling at the moment.

Best regards,
Alex

alexhauser:
Hi @kalaspuffar , 

I stripped out the common passwords list as requested, can you please have another look?

Thanks and best regards,
Alex

app/src/main/java/org/ea/sqrl/utils/PasswordStrengthMeter.java
```diff
@@ -0,0 +1,230 @@
...
+
+            result.PasswordLength = password.length();
+            if (result.PasswordLength <= 5) result.StrengthPoints = 0;
+            else if (result.PasswordLength <= 8) result.StrengthPoints = 1;
+            else if (result.PasswordLength <= 10) result.StrengthPoints = 2;
+            else if (result.PasswordLength <= 13) result.StrengthPoints = 5;
+            else if (result.PasswordLength <= 16) result.StrengthPoints = 10;
+            else result.StrengthPoints = 30;
+
+            for (int i=0; i < result.PasswordLength; i++) {
```
kalaspuffar: A micro-optimization would be to set booleans for the different classes and only check for the ones that you are missing and if all are set you don't loop through the rest.

app/src/main/java/org/ea/sqrl/utils/PasswordStrengthMeter.java
```diff
@@ -0,0 +1,230 @@
...
+
+                    case Good:
+                        txtPasswordStrength.setText(mContext.getText(R.string.password_strength_good));
+                        txtPasswordStrength.setBackgroundColor(
+                                ContextCompat.getColor(mContext, R.color.password_strength_good));
+                        break;
+                }
+
+                ImageView iv = null;
+                for (CharacterClass cc : result.CharClassCounts.keySet()) {
```
kalaspuffar: Again booleans would not require a loop and would only set colored drawables on the images directly. Less code and more clear.

app/src/main/java/org/ea/sqrl/utils/PasswordStrengthMeter.java
```diff
@@ -0,0 +1,230 @@
...
+            catch (Exception e) {
+                e.printStackTrace();
+            }
+        }
+    }
+
+
+    class PasswordStrengthResult {
+
+        public int StrengthPoints = 0;
```
kalaspuffar: In Java, we usually start variables, members, fields, and functions with lowercase.

I usually always use the get/set mentality instead of public fields just so you can override these later with functionality. But in this case, as it's an internal class I think it's ok to skip this.

app/src/main/res/layout/password_strength_meter.xml
```diff
@@ -0,0 +1,135 @@
...
+        android:layout_width="0dp"
+        android:layout_height="5dp"
+        android:max="9"
+        app:layout_constraintBottom_toTopOf="@+id/txtPasswordStrength"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/txtPasswordWarning" />
+
+    <TextView
+        android:id="@+id/txtPasswordStrength"
```
kalaspuffar: Not sure that we require some text to be all caps but if we want that design then I think it's more reasonable to add it to the text field instead of the requiring all translators to do this work.

android:textAllCaps="true"

app/src/main/java/org/ea/sqrl/utils/PasswordStrengthMeter.java
alexhauser: I wanted to be able to support displaying class counts, just as Steve does on his password haystack page, that's why I used ints. But I guess for our purposes, it wouldn't even be useful. I'll refactor the code to use booleans.

app/src/main/java/org/ea/sqrl/utils/PasswordStrengthMeter.java
alexhauser: Thanks for the input Daniel, very much appreciated!

I even considered doing it the "proper" way (get/set) but it just seemed so overly verbose for something that mundane.

As for Java coding conventions, my apologies for the occasional slip in that area, it's hard if you're mainly working in a different language and muscle memory kicks in.

app/src/main/res/layout/password_strength_meter.xml
alexhauser: I fully agree. Will be fixed in my next commit as well.

app/src/main/java/org/ea/sqrl/utils/PasswordStrengthMeter.java
alexhauser: Agreed.

alexhauser:
Hi @kalaspuffar ,

I've just committed the suggested changes, can you please have another look?

Thanks and best regards,
Alex

app/src/main/java/org/ea/sqrl/utils/PasswordStrengthMeter.java
```diff
@@ -0,0 +1,214 @@
...
+                }
+
+                if (result.lowercaseUsed) imgLowercase.setImageDrawable(ledGreen);
+                else imgLowercase.setImageDrawable(ledRed);
+
+                if (result.uppercaseUsed) imgUppercase.setImageDrawable(ledGreen);
+                else imgUppercase.setImageDrawable(ledRed);
+
+                if (result.digitsUsed) imgDigits.setImageDrawable(ledGreen);
+                else imgDigits.setImageDrawable(ledRed);
```
kalaspuffar: Again to keep to the coding standard I would either use brackets around the statements or use the online select like 

```result.digitsUsed ? ledGreen : ledRed```

File: app/src/main/java/org/ea/sqrl/utils/PasswordStrengthMeter.java [hunk 6]
```diff
@@ -0,0 +1,214 @@
...
+
+                char c = password.charAt(i);
+
+                if (c >= 65 && c <= 90) {       // Uppercase
+                    if (!result.uppercaseUsed) {
+                        result.strengthPoints += 2;
+                        result.uppercaseUsed = true;
+                    }
+                }
+                else if (c >= 97 && c <= 122) { // Lowercase
```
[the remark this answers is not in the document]
alexhauser: Point about coding style taken. Again, please bear with me in this regard.

I don't quite understand your point about evaluating the boolean first. As I am iterating over each character in the password, I have to determine the class of the current character first before making any other assumptions. I don't see any other possibility for optimization here, other than to stop if each class was used, which I'm already doing. Can you please explain in more detail or share some code?

app/src/main/java/org/ea/sqrl/utils/PasswordStrengthMeter.java
alexhauser: Point taken.

app/src/main/java/org/ea/sqrl/utils/PasswordStrengthMeter.java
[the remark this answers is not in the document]
kalaspuffar: Good point, my bad

kalaspuffar:
I will await the style changes and then merge.
If you like I can merge and change them if you like. It's small things but I want the code to have a single voice for readability.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

style fixes are applied in the latest commit, should now be ready to merge!

Best regards,
Alex

Issue #246, opened 2019-02-17
Stack Traces on Startup


sengsational:
*Short description of the issue*
Stack traces on startup of the app: `NoClassDefFoundError`
*Expected behavior*
Maybe just logging instead of stack traces?  Maybe this is not really a problem.  Or maybe it's just an artifact of how I've configured the project?
*Actual behavior*
When I started the app for the first time, the Logcat gets lots of NoClassDefFoundError stack traces (see here: https://github.com/sengsational/secure-quick-reliable-login/wiki/Logcat-with-NoClassDefFoundErrors).
*Steps to reproduce*
I forked the project, pulled the project into Android Studio 3.3.1, then ran it on my device "Moto G (5) Plus (Android 8.1.0, API 27)".

kalaspuffar:
Hi @sengsational 

Well, these errors will not be my main focus for a while. But as you have the Studio up and running if you like you may see if you can rectify them for us. Pull request are most welcome :)

Best regards
Daniel

sengsational:
It looks like if there's even a sniff of biometrics in an activity, the Android platform tries to exercise it, even though the method containing the code is never entered.

I found this [https://stackoverflow.com/questions/52896090/authenticationcallback-and-keygenparameterspec-crashing-app-despite-checking-api] which shows how in that case, if the biometric stuff is pulled out into a separate class, the problem is alleviated.   I'm not sure right now how to apply the knowledge to the code in SimplifiedActivity, but I'll think more about it.

sengsational:
I was able to get past the biometric stack traces by moving the biometric stuff into a new class only to realize that there was another problem " ClassNotFoundException OnUnhandledKeyEventListener", being discussed here: https://stackoverflow.com/a/52954286/897007 and here: https://issuetracker.google.com/issues/120750246

The good news is it's just junking up the very early bit of Logcat and the app works.

sengsational:
So there were two things causing stack traces in the Logcat on startup 1) Biometric stuff and 2) OnUnhandledKeyEventListener.

I found a solution to 2) here: https://stackoverflow.com/a/42957234/897007

So I moved the biometric stuff out into a separate class again (called it `LoginActivityHelper`).  At first, I tried to put it in the super class (`LoginBaseActivity`), but when I did that, the stack traces still were happening.  Only when it wasn't in a class that did not extend `Activity` does it remove the stack trace on startup.

[...]

PR #293, opened 2019-04-19
Remove stack traces on startup

sengsational:
Issue #246 (Stack Traces on Startup)

Description:
1) Added `configurations.all` section to Gradle build and
2) moved biometric code out of activity to a helper.

Changes:
Using information from https://stackoverflow.com/a/42957234/897007, I was able to remove the stack traces associated with `OnUnhandledKeyEventListener` when the app starts up.

By removing biometric related code from the `Activity` and putting it in a helper class, the stack traces on startup associated with biometrics no longer appear.

Discussion:
The goal was not to change the behavior of the application, just to eliminate the "noise" in logs.  I did have to change the method signature in one case to `public`, and also, I added a convenience method to get the handler (reduced the number of parameters passed).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -30,6 +30,17 @@ android {
     }
 }
 
+configurations.all {
+    resolutionStrategy.eachDependency { details ->
+        def requested = details.requested
+        if (requested.group == "com.android.support") {
+            if (!requested.name.startsWith("multidex")) {
+                details.useVersion "26.+"
+            }
+        }
+    }
+}
+
 dependencies {
     implementation fileTree(include: ['*.jar'], dir: 'libs')
     implementation 'com.android.support:appcompat-v7:28.0.0'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -7,8 +7,8 @@ android {
         applicationId "org.ea.sqrl"
         minSdkVersion 16
         targetSdkVersion 28
-        versionCode 40
-        versionName "0.14.0"
+        versionCode 41
+        versionName "0.14.1"
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
         vectorDrawables.useSupportLibrary = true
     }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -35,7 +35,7 @@ configurations.all {
         def requested = details.requested
         if (requested.group == "com.android.support") {
             if (!requested.name.startsWith("multidex")) {
-                details.useVersion "26.+"
+                details.useVersion "28.0.0"
             }
         }
     }
--- a/app/src/main/java/org/ea/sqrl/activites/LoginActivityHelper.java
+++ b/app/src/main/java/org/ea/sqrl/activites/LoginActivityHelper.java
@@ -1,0 +1,73 @@
+package org.ea.sqrl.activites;
+
+import android.hardware.biometrics.BiometricPrompt;
+import android.os.Build;
+import android.os.CancellationSignal;
+
+import org.ea.sqrl.R;
+import org.ea.sqrl.activites.base.LoginBaseActivity;
+import org.ea.sqrl.processors.BioAuthenticationCallback;
+import org.ea.sqrl.processors.CommunicationFlowHandler;
+import org.ea.sqrl.processors.SQRLStorage;
+
+import java.security.KeyStore;
+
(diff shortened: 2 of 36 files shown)


app/build.gradle
```diff
@@ -30,6 +30,17 @@ android {
     }
 }
 
+configurations.all {
+    resolutionStrategy.eachDependency { details ->
+        def requested = details.requested
+        if (requested.group == "com.android.support") {
+            if (!requested.name.startsWith("multidex")) {
+                details.useVersion "26.+"
```
kalaspuffar: The post you are referring to uses exact numbers when supplying version. So in our case, this should be 28.0.0 and nothing else.

Then again I don't really get how this change connects to the rest of the pull request.

app/src/main/java/org/ea/sqrl/activites/LoginActivityHelper.java
```diff
@@ -0,0 +1,73 @@
...
+import org.ea.sqrl.activites.base.LoginBaseActivity;
+import org.ea.sqrl.processors.BioAuthenticationCallback;
+import org.ea.sqrl.processors.CommunicationFlowHandler;
+import org.ea.sqrl.processors.SQRLStorage;
+
+import java.security.KeyStore;
+
+import javax.crypto.Cipher;
+
+public class LoginActivityHelper {
```
kalaspuffar: Why move a function used ones to a separate class when this case doesn't have a required version for running?

And why check for the version both at the call and the definition of the function?

app/build.gradle
sengsational: I copied and pasted that, so the "26.+" was not researched and if "28.0.0" makes sense, that can be changed.

As to how it relates to the pull request, it's that this bit of build definition prevents startup stack traces `OnUnhandledKeyEventListener`. 

So the pull request has two completely different "things", that each cause stack traces when the app starts.  This is just one of the two.

app/src/main/java/org/ea/sqrl/activites/LoginActivityHelper.java
sengsational: Good questions, both of them.  I struggled with both of those questions when I was working on it.

The separate class is just a weird thing about running the application.  There is no difference in the code logic, but if that bio-metric code is in a class that does not extend `Activity`, then the stack traces do not appear.  It's a work-around for what I consider a Google bug.

Checking the version in both places: Leaving the version check in the original place was for clarity.  Checking the version in the new class was in case the method was called from somewhere else.  Could be argued that it's confusing, or wasteful.  But I was going for safety.

app/build.gradle
kalaspuffar: Ok, only thing is that what I read this will override the settings you do in gradle.build and that would potentially give the implementation older versions.

kalaspuffar:
As I have limited time at the moment. Can you test and verify this pull request @alexhauser ?

alexhauser:
Hi @kalaspuffar and @sengsational ,

the change in build.gradle is a little beyond me, so I can't comment on that at the moment.

The code refactoring on the biometric login looks reasonable. I'd appreciate a comment on the helper class documenting its purpose (removing stack traces), so that anyone looking at it in 6 months will know why it's even there.

Also, regarding the duplicate version check, my feeling is that checking inside the class method is sufficient. The caller of the helper class shouldn't need to worry about this.

This is all just code review, I was not able to test this yet, but will do so as soon as time permits it.

Best regards,
Alex

kalaspuffar:
I concur with @alexhauser statement about the Gradle file. I don't understand that change and what it will entail. Does this mean that we need to update that value as well as the package versions. Why? Never seen a project using this construct. Is it an old requirement or does it still work? :/

alexhauser:
I was able to test the PR meanwhile and can confirm that the biometric login still works as expected.

@sengsational , since we are all unsure about the `build.gradle` part, I would suggest to remove this code for now. I would prefer having some weird stack traces over actively adding code that we don't fully understand the possible consequences of. What do you think?

Best regards,
Alex

sengsational:
@alexhauser & @kalaspuffar ,
The stack traces are just a minor annoyance and we can abandon these changes.  I suspect that "some day", these problems will be fixed by the Android platform and at that point, they'd be extra baggage in the code base.

Issue #298, opened 2019-04-30
Consolidate rescue code input layout


alexhauser:
### Optimization suggestion

This applies to this particular part of our UI:

![rescue_code_input](https://user-images.githubusercontent.com/4005543/56957203-ab758800-6b46-11e9-821f-a2a3de5a0f4c.JPG)

I noticed that we have the same (or at least very similar) layout code for entering the rescue code 
duplicated on all of these activities:
```
activity_enable_account.xml
activity_rekey_verify.xml
activity_remove_account.xml
activity_rescuecode_show.xml
activity_rescuecode_enter.xml
activity_reset_password.xml
```
Does this serve any special purpose or are there important differences between the different layouts that I didn't notice on my quick review? If no, should we consolidate this to a separate .xml layout file which we can then include in all of these activities?

This would make the layout code much more maintainable imho.

Thanks in advance for any input or suggestions!

kalaspuffar:
Hi @alexhauser 

Not to my knowledge. Been a while since I wrote that code thought :)

Consolidation is good. Most of the code have been written in a ad-hoc manor knowing I had to change it later. This part might have stabilized enough to break out though.

Best regards
Daniel

alexhauser:
Thanks for sharing the background info, @kalaspuffar ! I'm starting to work on this.

Best regards,
Alex

PR #303, opened 2019-05-01
Consolidate rescue code layout and code

alexhauser:
Issue #298 .

Description:
Removed duplicate layout code and replaced it with only one import file: *rescue_code_input.xml*
Also added a helper class for dealing with input verification. The code of the affected activities was also refactored to use the this new helper class. All the logic is now in one place.

The following activities (layout and code) were affected:
```
activity_enable_account.xml
activity_rekey_verify.xml
activity_remove_account.xml
activity_rescuecode_show.xml
activity_rescuecode_enter.xml
activity_reset_password.xml
```

Changes/improvements:
- Removed duplicate layout code
- Removed duplicate verification code
- Removed error popups in favour of on-the-fly form input checks. Also, the "next"-button is now enabled/disabled on the fly for most of the activities.
- Focus now jumps to the next field automatically
- Resolved a display bug in activity_rescuecode_show.xml which would let the buttons move off screen if the description text was long (like with the German translation)

I have also found some other bugs and glitches along the way which I will address separately.

Looking forward to your review, @kalaspuffar and @sengsational !

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -59,7 +59,7 @@
 
         <activity android:name=".activites.create.RescueCodeShowActivity"
             android:label="@string/title_rescuecode_show"
-            android:windowSoftInputMode="stateHidden"
+            android:windowSoftInputMode="stateAlwaysHidden"
             />
 
         <activity android:name=".activites.create.RescueCodeEnterActivity"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -59,6 +59,7 @@
 
         <activity android:name=".activites.create.RescueCodeShowActivity"
             android:label="@string/title_rescuecode_show"
+            android:windowSoftInputMode="stateHidden"
             />
 
         <activity android:name=".activites.create.RescueCodeEnterActivity"
--- a/app/src/main/res/layout-land/activity_rescuecode_show.xml
+++ b/app/src/main/res/layout-land/activity_rescuecode_show.xml
@@ -7,111 +7,30 @@
     android:id="@+id/rescueCodeShowActivityView"
     tools:context="org.ea.sqrl.activites.create.RescueCodeShowActivity">
 
-    <android.support.constraint.ConstraintLayout
-        android:id="@+id/inputRow1"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:layout_marginEnd="8dp"
-        android:layout_marginStart="8dp"
+    <TextView
+        android:id="@+id/txtRescueCodeShowDescription"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        android:layout_marginStart="16dp"
         android:layout_marginTop="16dp"
-        android:focusable="false"
-        android:layoutDirection="ltr"
+        android:layout_marginEnd="16dp"
+        android:text="@string/rescuecode_show_desc"
+        app:layout_constraintBottom_toTopOf="@+id/rescueCodeInput"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@+id/textView17">
+        app:layout_constraintTop_toTopOf="parent" />
 
-        <TextView
-            android:id="@+id/txtRecoverCode1"
-            android:layout_width="72dp"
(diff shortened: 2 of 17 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -59,6 +59,7 @@
 
         <activity android:name=".activites.create.RescueCodeShowActivity"
             android:label="@string/title_rescuecode_show"
+            android:windowSoftInputMode="stateHidden"
```
kalaspuffar: I wonder if we don't want the stateAlwaysHidden instead for this activity. I'm not sure we could navigate back to it today but maybe that would be an issue in the future.

"stateAlwaysHidden" | The soft keyboard is always hidden when the activity's main window has input focus.

app/src/main/res/layout-land/activity_rescuecode_show.xml
```diff
@@ -7,111 +7,30 @@
...
-        android:layout_marginEnd="8dp"
-        android:layout_marginStart="8dp"
+    <TextView
+        android:id="@+id/txtRescueCodeShowDescription"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        android:layout_marginStart="16dp"
         android:layout_marginTop="16dp"
-        android:focusable="false"
-        android:layoutDirection="ltr"
```
kalaspuffar: Why did we remove these?

I'm not sure if the text was focusable for some reason but that might have been a change to handle accessibility. the layoutDirection is strange though as we have arabic translation that should be able to be rtl :/

app/src/main/res/layout/activity_rescuecode_enter.xml
```diff
@@ -7,117 +7,28 @@
...
+    <TextView
+        android:id="@+id/rescueCodeEnterDescription"
+        android:layout_width="0dp"
         android:layout_height="wrap_content"
-        android:layout_marginEnd="8dp"
-        android:layout_marginStart="8dp"
+        android:layout_marginEnd="16dp"
+        android:layout_marginStart="16dp"
         android:layout_marginTop="16dp"
-        android:layoutDirection="ltr"
```
kalaspuffar: See question above.

kalaspuffar:
Hi @alexhauser 

Did you move the show information to the bottom of the display? Seems a bit strange on larger devices.

issue2

Best regards
Daniel

app/src/main/res/layout-land/activity_rescuecode_show.xml
alexhauser: I think this is only the diff tool screwing up. It's not comparing the same components here.
```
android:layoutDirection="ltr"
android:focusable="false"
```
This code belonges to the `ConstraintLayout` holding the rescue code input and is still present, but was moved over to `rescue_code_input.xml`. There is a `TextView` now in `activity_rescuecode_show.xml` where the `ConstraintLayout` was before, which is of course missing these attributes. So all should be fine. Can you confirm?

app/src/main/res/layout/activity_rescuecode_enter.xml
alexhauser: Same as above I guess.

app/src/main/res/layout-land/activity_rescuecode_show.xml
kalaspuffar: Correkt

alexhauser:
> Hi @alexhauser
> 
> Did you move the show information to the bottom of the display? Seems a bit strange on larger devices.

Hi @kalaspuffar ,

yes, I did this mainly for portrait mode where I needed the description text to take the rest of the screen space and be scrollable without pushing the rescue code and the buttons off screen. This was the case on the German translation before this chage.

I agree that on big devices it might be a bit odd in landscape mode, but I just wanted to be on the safe side so that things don't get pushed off screen on smaller resolutions.

app/src/main/AndroidManifest.xml
alexhauser: I agree. I'll change it to `stateAlwaysHidden`.
Thanks for taking time to review! :)

kalaspuffar: Hi @alexhauser 

Thank you for giving your time to this repository. And helping me with code review and testing on my pull requests.

The quality of this project is important and if we can help each other so we find issues early then that's a win-win for everyone.

Best regards
Daniel

Issue #301, opened 2019-05-01
Same device alternative id don't work.


kalaspuffar:
### Issue or error solving.

The alternative id dialog doesn't work as expected when the same device login is used.

1. Click on a link on the device.
2. Select option
3. Select alternative id
4. Fill in the ID and password

Expected:
The user is returned to the site and is logged in.

Actual behavior:
The user gets stuck on the first login page and needs to press back in order to return to the site.

PR #304, opened 2019-05-02
Return to webpage when you use alternative id function.

kalaspuffar:
Issue #301.

Description:
The application should return to the website when you use the alternative ID function.

Changes:
Added calls to close all the extra views when entering the optional features of the login window.


kalaspuffar:

Can you review and test this pull request?

Best regards
Daniel

Issue #306, opened 2019-05-02
Strength meter in dark mode.


kalaspuffar:
### Issue or error solving.

Strength meter error message is unreadable in dark mode.
![issue1](https://user-images.githubusercontent.com/1424084/57105750-de2ca580-6d2b-11e9-90f2-54ad3a552a88.png)

kalaspuffar:
Hi @alexhauser 

Can you verify dark mode functionallity.

You should be able to set it with 
AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_YES);

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

thanks for pointing this out. I've actually noticed this myself today and have already prepared a fix in my local repository. Do you agree that we should use a fixed font color on those components that suits all three states? Or is there a better way to solve this? 

Thanks and best regards,
Alex

kalaspuffar:
Hi @alexhauser 

Not really. Think of dark mode as a view used by light sensitive people or when using the app in the dark. So finding colors and fonts that work and follow the overall asterisks of the application is more important.

Then again we might get some more input about flows and color scheme when the design team have looked at the application further.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

the dark mode glitch should be fixed with the last PR, however, I've noticed another display bug: When translations are long, the character class indicators get pushed off the screen to the right. Guess I'll open a new issue for this tomorrow.

Good night, everyone ;)
Alex

PR #307, opened 2019-05-02
Fix strength meter dark mode and display issues

alexhauser:
Issue #306.

Description:
Strength meter font was unreadable in dark mode.

Changes:
Set fixed font color for strength meter components.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -39,6 +39,7 @@ dependencies {
     implementation 'com.journeyapps:zxing-android-embedded:3.2.0@aar'
     implementation 'com.google.zxing:core:3.2.1'
     implementation 'com.github.joshjdevl.libsodiumjni:libsodium-jni-aar:2.0.1'
+    implementation 'com.google.android:flexbox:1.0.0'
     testImplementation 'junit:junit:4.12'
     androidTestImplementation 'com.android.support.test:runner:1.0.2'
     androidTestImplementation 'com.android.support.test:rules:1.0.2'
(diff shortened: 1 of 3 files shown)


alexhauser:
Hi @kalaspuffar , please wait with merging this, I'll also add a commit for the layout bugfix (issue #308) shortly. Thanks.

alexhauser:
Ok, display bug fix is now included. Review would be appreciated, thanks.

kalaspuffar:
Hi @alexhauser 

I've looked at the pull request and the main issue I have with this one is the extra dependency. Adding extra libraries for layout seems strange. Could we do the same things with the constrained layout? Or perhaps a linear layout?

Best regards
Daniel

kalaspuffar:
Hi again.

I think that if we can find a good solution to this one then I want to build a release. We have a good amount of new stuff that we want to push.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

i know it's rediculous, but apparently neither `ConstraintLayout` nor `LinearLayout` nor any other "built-in" `ViewGroup` containers support wrapping its children to the next line. At least that's the conclusion I ended up with researching a solution. Anybody please correct me if I'm wrong.

It seems the only two options here are either pulling in Google's FlexBox layout or writing our own ViewGroup-derived layout class.

I opted for using the well-tested and proven Google solution instaed of rolling my own. But if you prefer writing our own `ViewGroup` class and getting rid of the dependency, I can certainly do so. 

Please let me know how to proceed.

Thanks and best regards,
Alex

Issue #260, opened 2019-03-10
Handle .sqrl file intent


kalaspuffar:
Using file intent and the import activity to import identity files into application.

alexhauser:
Hi @kalaspuffar , 

is this something you're actively working on? Otherwise I could offer some help here as well.

Best regards,
Alex

kalaspuffar:
Hi @alexhauser 

Nope, I tried but have not had time to look into this yet. Tried a few different mime-types but never got it to trigger. If I find how to trigger the share function it should be trivial to enable import.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

I tried to get it to trigger for the .sqrl file extension only, and indeed this seems to be impossible to do, at least on anything >= Nougat:

In the SO question https://stackoverflow.com/questions/3760276/android-intent-filter-associate-app-with-file-extension/12915288 someone posts:

> So from what I've seen, tried and read, Android simply cannot distinguish between file-extensions and the pathPattern tag is basicly a gigantic waste of time and energy. If you are fortunate enough to only need files of a certain mime-type (say text, video or audio), you can use an intent-filter with a mime-type. If you need a specific file-extension or a mime-type not known by Android however then you're out of luck.

And another one:

> Android 7.0 Nougat no longer permits file sharing between apps using a file URI. A content URI must be used. However, a content URI does not allow a file path to be shared, only a mime type. So you cannot use a content URI to associate your app with your own file extension.

Which leaves us with mime types only I guess. And since the mime type for our .sqrl files is probably undefined or not known to Android, it looks like it's "all or nothing" here. The intent-filter does trigger if we don't specify any special mime type, but this means SQRL will be offered for *any* file type that is being shared. Yuck!

Best regards,
Alex

kalaspuffar:
Hi @alexhauser 

Yes, that is not a good solution or a nice solution but if we only have one we might need to support it for now. But we need to check the file if it has a header or else show a nice error message so users aren't confused by this feature.

Best regards
Daniel

PR #313, opened 2019-05-09
Handle .sqrl file intent

alexhauser:
Issue #260.

Description:
Enable .sqrl identity file import by choosing the SQRL app in the "*open with app*" android dialog after tapping on a .sqrl file in downloads or a file explorer app.

Changes:
Register an intent filter in the manifest for `ImportActivity` to receive file `VIEW` intents for the mime types `text/*` and `application/*`. Detect such an intent in `onCreate()` and handle it the same way as the QR import would. In case of an invalid file, all buttons are disabled and an error message is displayed in the activity's description TextView.

I guess we'll see how this works on different APIs/devices, since there apparently were a lot of changes in this particular Android API over time and it's a BIG MESS!! Maybe we'll have to broaden the mime types in the future.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/utils/Utils.java
+++ b/app/src/main/java/org/ea/sqrl/utils/Utils.java
@@ -15,6 +15,7 @@ import com.google.zxing.FormatException;
 import org.ea.sqrl.database.IdentityDBHelper;
 import org.ea.sqrl.processors.SQRLStorage;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Locale;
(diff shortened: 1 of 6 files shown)


alexhauser:
Hi @kalaspuffar ,

from issue #314:

> About #260 I might have found a bug in it or the file export. Could not import with password when testing in emulator. Could be an emulator issue and I did it at work between tasks so have not tested it extensively.

Interesting. I've successfully tested the import feature with several import files. Waiting for further input on this one.

Thanks and best regards,
Alex

app/src/main/java/org/ea/sqrl/utils/Utils.java
```diff
@@ -76,4 +79,20 @@ public static void refreshStorageFromDb(Activity activity) throws Exception {
         SQRLStorage sqrlStorage = SQRLStorage.getInstance(activity);
         sqrlStorage.read(identityData);
     }
+
+    public static byte[] getFileIntentContent(Context context, Uri contentUri) {
+        if (contentUri == null) return null;
+
+        try {
+            InputStream inputStream = context.getContentResolver().openInputStream(contentUri);
+            byte[] data = new byte[inputStream.available()];
```
kalaspuffar: "Note that while some implementations of InputStream will return the total number of bytes in the stream, many will not. It is never correct to use the return value of this method to allocate a buffer intended to hold all data in this stream."

https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#available()

Can you use a ByteBuffer or a ByteArrayOutputStream instead so we can ensure that all bytes are read?

alexhauser: Correct. I read that `InputStream.available()` will return the full number of bytes if reading from a file off of local storage, but the docs make it absolutely clear that this is never a correct implementation.   I will fix this asap.

alexhauser: Should be fixed with latest commit.

Issue #314, opened 2019-05-11
Intent.ACTION_OPEN_DOCUMENT


kalaspuffar:
Research feature

alexhauser:
Hi @kalaspuffar ,

sounds like a great feature. This would enable easy import of .sqrl files from local storage as well as from cloud providers, usb devices etc. directly from the app. Also, it would be a great complementary feature to #260 and should play nicely with my current implementation (pending PR) thereof.

And allthough the docs would want us to use `ACTION_GET_CONTENT` instead...
> ACTION_OPEN_DOCUMENT is not intended to be a replacement for ACTION_GET_CONTENT. The one you should use depends on the needs of your app:

> - Use ACTION_GET_CONTENT if you want your app to simply read/import data. With this approach, the app imports a copy of the data, such as an image file.

> - Use ACTION_OPEN_DOCUMENT if you want your app to have long term, persistent access to documents owned by a document provider. An example would be a photo-editing app that lets users edit images stored in a document provider.

... I still think having all the document provider sources in one single UI is compelling.

One question I have is how would we surface this feature? In `advanced functions` -> `Import`, should we present another choice where one can choose between "*Scan identity QR code*" and "*Import identitiy from file*"?

Best regards,
Alex

kalaspuffar:
Hi @alexhauser 

All your comments above align with what I've been thinking since I heard about the feature.
Listening to all Google I/O talks at the moment. Interesting that this feature haven't been more highlighted.

Is it supported in all of our supported API versions?

Best regards
Daniel

kalaspuffar:
Hi again @alexhauser 

Maybe you should do this new feature with moving things around and adding GET_CONTENT in a separate branch/PR to make the review process quicker.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

great to hear that we're on the same page on this one! 

> Is it supported in all of our supported API versions?

Not exactly, it is available starting from API level 19. But we could implement `ACTION_OPEN_DOCUMENT` for >= 19 and `ACTION_GET_CONTENT` for earlier APIs.

> Maybe you should do this new feature with moving things around and adding GET_CONTENT in a separate branch/PR to make the review process quicker.

I fully agree. But then i'd like to wait for the PR for #260 to be reviewed/merged, so that I can base everything off of a clean, reviewed state.

Best regards,
Alex

kalaspuffar:
Hi @alexhauser 

[...]

PR #316, opened 2019-05-15
File import feature and import UI consolidation

alexhauser:
Issue #314.

Description:

Enable importing a sqrl identity file from within the app. Also, consolidate the import UI.

Changes:
- Implement `Intent.ACTION_GET_CONTENT` to pick an identity file for import
- Add `ImportOptionsActivity` which now hosts all currently available import-options:
  - Scan identity qr code
  - Import identity file
  - Import with text input

import_options

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -117,6 +117,10 @@
             android:label="@string/title_export_options"
             />
 
+        <activity android:name=".activites.identity.ImportOptionsActivity"
+            android:label="@string/title_import_options"
+            />
+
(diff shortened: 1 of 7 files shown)


kalaspuffar:
Hi @alexhauser 

At the moment the accessibility tests don't work because of a bug in the testing framework but when we add new views we should also add tests to ensure that we keep an accessible GUI.

I will merge this pull request now as the code is good as is but in the future you could keep it in mind. The tests are located in ```app\src\androidTest\java\org\ea\sqrl\AccessibilityInstrumentedTest.java```

Best regards
Daniel

Issue #325, opened 2019-05-30
Wrong password error message covered by logon popup


sengsational:
*Short description of the issue*
Message `showErrorMessage(R.string.decrypt_identity_fail);` is covered by login popup.

*Expected behavior*
The login popup does not cover the error.

*Actual behavior*
The login popup covers the error.

*Steps to reproduce*
Scan a web login QR code.
Type the wrong password.
The login popup shows.
Dismiss the login popup without using it.
Now you can see the error message.

*Discussion*
I thought we had fixed this at one time, didn't we?

kalaspuffar:
Also thought we solved it. Maybe some code has reverted during all refactoring.

alexhauser:
This one can be closed as well, right?

PR #326, opened 2019-05-31
remove login popup after failed password

sengsational:
Issue #325 

Description:
Removed the login popup after a failed login attempt.

Changes:
If the user types the wrong password, they will see the error message.  Once they dismiss the error message, the login popup will show as the next action.


alexhauser:
I've just tested the fix and everything seems to be working as intended.
Nice catch, @sengsational !

Issue #317, opened 2019-05-20
Improve user experience of MainActivity


alexhauser:
### Enhancement: Improve user experience

I am aware that we are awaiting input from UI/UX experts in the forums at the moment, but as there hasn't been any real progress over the last few weeks, I thought I would ask if we shouldn't at least improve what we've got so far.

To me, the most "frightening" part of the UI at the moment is what I like to think of as the *"button graveyard"*, which is our `MainActivity`, aka *"Advanced operations"*. At the moment, we have **11 buttons** stacked up vertically, with button topics mixed up wildly, which I think everyone will agree, is not the best user experience one can think of.

One currently still open PR will already consolidate the import options into its own activity which helps a little. But there is much more room for improvement here I think.

Here are some suggestions for quickly implementable fixes, which, while not providing a complete solution to the overall UX problem, would at least mitigate it a little:

- [x] **Remove the button `Use Identity`.** 
Am I right in assuming that this a remnant from the time before `SimplifiedActivity`? To me, it is not even immediately clear what "use identity" even means, and I think we should keep the operational aspects of the app separated from the "identity management" aspect. I think the main action in `SimplifiedActivity` should be the only way of invoking a qr scan.
- [x] **Move `Settings` into the main overflow menu**
We could even think about showing it as an icon in the action bar. But I think it makes sense moving it one level "higher", away from the other purely "identity management" related options.
- [x] **Move `Change Password`, `Reset Password` and `Rekey` into its own sub activity.**
This could then be called `PasswordOptions` or something similar.
- [x] **Consolidate import options** 
Already implemented in pending PR *"File import feature and import UI consolidation"*
- [x] **Rename `MainActivity`** 

This would leave us with "only" the following options in `MainActivity`:

 - Create New Identity
 - Remove Identity
 - Rename Identity
 - Export Identity
 - Import Identity
 - Password/Rekey Options

I also think it would then be appropriate to rename `MainActivity` to `IdentityManagementActivity` and call the correspoding overflow menu item `Identity Management` instead of `Advanced operations`. 

@kalaspuffar ,
I would totally understand if you'd rather wait for the UI/UX experts to finally present something, but this is bugging me already for quite some time and I wanted to at least get a discussion going.

Looking foward to your feedback, guys!

Best regards,
Alex

kalaspuffar:
Hi @alexhauser 

The first easy issue is the name change. And I agree we the names you suggested are better.

[...]

PR #327, opened 2019-05-31
Improve IdentityManagementActivity

alexhauser:
Issue #317.

Description:

Further improve the user experience of `IdentityManagementActivity` (formerly known as `MainActivity`) as discussed in the corresponding issue.

Changes:

- Remove all buttons that correspond to actions performed on the currently selected identity from the activity
- Add a settings icon next to the identity spinner
- Move the functionality of the removed buttons into a popup menu which is accessible through the new settings icon

This only leaves the buttons `Create new` and `Import` visible directly.
The activity should look and feel drastically more lightweight now:

The popup menu now hosts all the options related to the current identity:

Selecting "Password options" will spawn another popup menu in-place:

It is debatable whether the rekey functionality should be hosted in the "password options" sub-menu or moved one level up to become its own action in the main popup menu.

As usual, any feedback is highly appreciated.


alexhauser:
Sorry, failed merge, will repost.

PR #336, opened 2019-06-04
Better contrasting colors for night and day mode.

kalaspuffar:
Description:
Look at a Google IO talk about Material design suggesting using 700 and 200 for the colors of different layouts to have good contrasts

Changes:
Use the same color for primaryColor and primaryColorDark so the status bar has the same color as our appbar. Also changed the accentColor to use two new purple colors with the right contrast difference.

What do you think about the new colors? @alexhauser @sengsational @sesam

### DIFF
--- a/app/src/main/res/values-night/colors.xml
+++ b/app/src/main/res/values-night/colors.xml
@@ -2,7 +2,7 @@
 <resources xmlns:tools="http://schemas.android.com/tools">
     <color name="colorPrimary">#303f9f</color>
     <color name="colorPrimaryDark">#303f9f</color>
-    <color name="colorAccent">#7B1FA2</color>
+    <color name="colorAccent">#CE93D8</color>
 
 
     <color name="black_overlay">#66000000</color>


alexhauser:
Hi @kalaspuffar ,

first of all, I think that defining a good color scheme is vital and I'm happy that you're tackling this challenge. 

I've just checked out your PR and I have to admit that I'm a bit ambivalent about the new colors. On one hand, I get what we're trying to accomplish, but on the other hand the new colors just feel a little too "harsh" to me. 

Maybe it's just because I'm so used to the old colors and this will only take some time to get used to, I don't know.

What I can say right now is that I definitely like the "day" mode better than the "night" mode. The softish purple to me looks a little strange in night mode, especially on the big button in `SimplifiedActivity`.

I think I'll sleep over it and have look at it again tomorrow.

Best regards,
Alex

sengsational:
I noticed that we have abc_*_text.... in the res/color/ folder, then we have another location res/values/colors, and the xml files there (at least in Android view..I know it's different in the file system).  That might get consolidated with any color work that gets done.

I like the dark purple color, but it seems weird to have dark purple in the day and light purple at night.  It seems to me that more than having a different accent color, the background would be more important. But I'm not sure how to change to night mode, so maybe the white background is addressed outside of the two color files, which appear to be the same except for `colorAccent`.

kalaspuffar:
Hi again @alexhauser and @sengsational 

The talk I took inspiration from was https://www.youtube.com/watch?v=9NDLR3COU7Y if you want to know how they thought about accenting colors and contrast. I'm not saying that this is the final look but I'm leaning towards having brighter colors to accent in dark mode so the black can contrast better.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar, 

thanks for sharing the video. I looked at it and found it interesting and educational.

And while I'm still not completely sure about the choice of colors, I think it is important to address the day/night mode topic and get started on this. That's why I'd opt for merging this now to get it out into public and get feedback. We can always tweak the colors later if we think it's necessary.

Best regards,
Alex

alexhauser:
@sengsational 
> I noticed that we have abc_*_text.... in the res/color/ folder, then we have another location res/values/colors, and the xml files there (at least in Android view..I know it's different in the file system). That might get consolidated with any color work that gets done.

I guess leaving both of those is legit:
https://stackoverflow.com/questions/16951960/difference-between-res-color-and-res-values-colors-xml-in-android-resources-fold

> In very general form, you can refer /values as a fix color, eg: black = #000000, so you can call "black" in layout instead of writing "#000000". While /color gives you the option to have different color on different view stat, eg: different color on disabled and enabled button. – Chor Wai Chun Jun 6 '13 at 1:53

sengsational:
Thanks for that research and explanation!

Issue #330, opened 2019-05-31
Idle timeout setting overflow


sengsational:
### Issue: The user setting for idle timeout can be too large

*Short description of the issue*
The user can enter a large value into idle timeout, then the field will show a different value if the identity block is re-read.

*Expected behavior*
The user would not be permitted to enter a value that's beyond the area to store it (two bytes), and so the value they type would be consistent.

*Actual behavior*
The user can type a value that goes beyond the capability of the underlying storage technique, so the value may change.  Also, it might cause a negative timeout value.

*Steps to reproduce*
Go into settings and type something like 999999999
Save and exit settings
Go back into settings.  See that 999999999 is no longer displayed

*Discussion*
I found this because I noticed that after re-installing and re-configuring the timeout to a large value, my QuickPass was being wiped out immediately.

PR #341, opened 2019-06-06
Prevent extreme idle timeout values

sengsational:
Issue #330.

Description:
If large value entered for idle timeout, guidance appears and also limits the value.

Changes:
Added listener on idle timeout field in the settings activity.  Expanded the error dialog to show something other than the heading of "...error".  I selected the word "guidance".


app/src/main/java/org/ea/sqrl/activites/SettingsActivity.java
```diff
@@ -60,6 +64,29 @@ protected void onCreate(Bundle savedInstanceState) {
         btnSettingsSave.setOnClickListener(v ->
             savePopupWindow.showAtLocation(savePopupWindow.getContentView(), Gravity.CENTER, 0, 0)
         );
+
+        txtSettingsIdleTimeout.addTextChangedListener(new TextWatcher() {
+            @Override
+            public void afterTextChanged(Editable editable) {
+                if (editable.length() > 2) {
+                    Log.v("sengsational", "after " + editable.length());
```
alexhauser: I don't know how we want to handle logging in general, but I don't see any verbose logging elsewhere in the app. The tag "sengsational" suggests that this was for debugging only. Should this be removed?

app/src/main/res/values/strings.xml
```diff
@@ -250,4 +250,6 @@
 <string name="title_identity_management">SQRL - Identity Management</string>
 <string name="language_chinese">Chinese</string>
 <string name="action_password_options">Password options</string>
+<string name="idle_timeout_guidance">So you\'ll be able to remind yourself periodically of your SQRL master password, we only let you go one week (%1$s minutes), then you\'ll have to type it again.</string>
```
alexhauser: I suggest to avoid words like "we", which could raise questions with our users (like *who exactly is "we"?*), and go for a more neutral form like for example

`So you don't forget your SQRL master password, the maximum QuickPass timeout is one week ( %1$s minutes), after which you\'ll have to type in the master password again.`

alexhauser:
Good work Dale, I like how you don't just throw a random error message but give some guidance to the user.

As for translation, @kalaspuffar what is the best practice here? @sengsational adds the english strings to all the languages. Is this correct or should we just fill the default `strings.xml` and leave the other languages to POEditor?

Also, I think we should use the opportunity to rename `Idle timeout` to `QuickPass timeout`. I always thought "Idle timeout" was very nondescript and should be changed. What do you guys think?

Best regards,
Alex

sengsational:
> As for translation, @kalaspuffar what is the best practice here? @sengsational adds the english strings to all the languages. Is this correct or should we just fill the default `strings.xml` and leave the other languages to POEditor?

Maybe there's some configuration magic to be done in my Android Studio, but when I tried to just add the text to the default `strings.xml`, the compiler complained.  So I *had* to paste the English everywhere.  I'm glad for the tool in Android Studio that let's me "paste/tab/paste/tab..." through all of the languages.  

> Also, I think we should use the opportunity to rename `Idle timeout` to `QuickPass timeout`. I always thought "Idle timeout" was very nondescript and should be changed. What do you guys think?

I agree.  Should I do this with this PR?

app/src/main/res/values/strings.xml
sengsational: I like yours better.  Mine sounds like I'm typing an email to my friend, hehe!

app/src/main/java/org/ea/sqrl/activites/SettingsActivity.java
sengsational: Oops!  I meant to clean this up.  I put a unique tag on it so I can find it, hehe!

As for logging and comments, this code is devoid of any of those, for the most part.  When I code my own stuff, I put tons of logging and I leave it in the code base because when the code is built for `release`, it's not included, so doesn't bog down the devices of the "real" users.  A matter of taste, I suppose.  But I've always tried to bend my techniques to match the existing code, so if I remember, I pull out the logging.

alexhauser:
I hope @kalaspuffar will shed some light on the questions raised above.

kalaspuffar:

Sorry for the late reply, I've been off work to celebrate my wife now being a nurse.

When it comes to the strings it's fine to add them to one language or multiple. The correct approach is to create another string on POEditor and that might not be something everyone has access to do. But I will update POEditor as fast as possible. Then the script I built will fetch the strings from POEditor and update the Android build. 

I tend to do that regularly and when I release. If you want you could highlight the added strings so I easily could copy paste them into POEditor.

Best regards
Daniel

kalaspuffar:
Hi @sengsational 

I've merged the other pull requests in the list. I would like to merge this one before I release the next release. At the moment we have merge conflicts with the message strings.

I've added all strings to POEditor and also updated master with the new strings. So removing message files should resolve the issue. Or you may resolve the conflicts if you like.

Either way, I'll wait for your last touch ups on this request and then merge and build a release, I would like to do it tomorrow if possible. We have a lot of nice features that I want to deliver. :)

Best regards
Daniel

sengsational:
I think I understood the *concept* of removing the message files from the branch, but I was timid about doing that, so I just did a diff between mine and what were in origin/master.  So dd2f31b should have all `string.xml` to be the same as what's in master.

kalaspuffar:
Hi @sengsational 

It's fine. I'll merge and then reset all the strings from the translations at POEditor :)
When it comes to translations POEditor is always the truth.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

it seems that some changes in `strings.xml` (changing all occurrances of *"Idle timeout"* to *"QuickPass timeout"*) have been lost while merging. Can you please bring them back? I really think "idle timeout" is a very nondescript term that we should not be using.

Thanks, Alex

PR #346, opened 2019-06-10
Update README.md

sesam:
Description:
Adds an above-the-fold quick start for users who wander in, to capture their interest before presenting them with the long-form prose.
(Optimally we'd also add some fancy-looking graphics, probably a GIF or similar animation, to show off how nice logging in with SQRL Login is!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,26 @@
 
 [![Build Status](https://travis-ci.org/kalaspuffar/secure-quick-reliable-login.svg?branch=master)](https://travis-ci.org/kalaspuffar/secure-quick-reliable-login)
 
-This repository is an implementation for SQRL (Secure Quick Reliable Login) on Android.
+# What
+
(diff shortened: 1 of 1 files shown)


kalaspuffar:
Some information is duplicated, and to make this over the fold information; we need to identify the audience. At the moment I've optimized this page for developers. Users will find the application first from the forum and other places. So this will be the last page they visit after already using the app.

alexhauser:
Hi Simon, I agree with Daniel that your suggested changes are targeted towards an audience of end users that we don't typically see here on GitHub. I would prefer sticking with the developer oriented content here.

Once we have something end user facing, your suggestions will become valuable.

sesam:
Great 👍  I'll be happy to make the github page more geared towards capturing first-time visitors at some other time, if you want, and I can also take directions to work from. Closing this for now, so the PR doesn't drag on.

PR #353, opened 2019-06-20
Fix for long data

kalaspuffar:
Description:
I ran across this little gem when I was testing the new SQRL Login WordPress plugin. We handle long urls and data that stretches more than one segment incorrectly. Interesting that this bug could be left this long, I left it as a... I fix that later :)

Changes:
Added an interation over all the segments in the QRCode to get all data. Also fetched the scan result instead when we want the string data.

If you want to review and comment it's appreciated @alexhauser and @sengsational


app/src/main/java/org/ea/sqrl/utils/Utils.java
```diff
@@ -37,16 +37,18 @@
      * @return  The string without any extra information.
      */
     public static byte[] readSQRLQRCode(Intent data) throws FormatException {
-        return data.getByteArrayExtra("SCAN_RESULT_BYTE_SEGMENTS_0");
+        byte[] qrCode = new byte[0];
+        for(int i=0; i<10; i++) {
```
alexhauser: Is the number 10 for the number of segments chosen arbitrarily or is a number specified by a specification? If it's the first, I would suggest to make this a `for (int i=0; ; i++)` and drop out of the loop only if `newSegment == null` .

kalaspuffar: Hi Alex.

Well, I think there are only 5 segments, but I did not find documentation actually telling you how many there could be. But there is an upper limit on how large a QRCode can be and that prohibits many segments. And an SQRL identity will probably never be larger than 2-3 segments if we look at the current implementation.

I don't like loops that loop forever. I had a lot of bad bugs that locked devices or programs due to these kinds of loops. So can I set an upper limit I like to have one.

Then again this might be a case we can allow it as an array probably can't have infinite text strings.

Best regards
Daniel

kalaspuffar:
Changed as per your suggestion. Give this comment a thumbs up and I will merge.

Issue #354, opened 2019-06-21
Include info / help icons


m2-dev:
### Feature request.

Provide an info icon throughout settings/help dialogues to explain terminology and reasons for/against using various features

alexhauser:
Jfi, I've started working on this.

PR #356, opened 2019-06-24
Info/help icons for identity settings

alexhauser:
Issue #354.

Description:

This will add info icons to the identity settings, which will show some descriptive help text in a dialog popup when tapped/clicked. Also, an `IdentitySelector` was added so that one can easily see for which identity the displayed settings are valid (and to easily change identity of course).

Changes:

- Included info icons that when being tapped will show some help text related to the current setting
- Renamed _SettingsActivity_ to _IdentitySettingsActivity_ to make it clear that these settings are identity specific and also in preparation of supporting application wide settings at some point which could then get the more general name
- Changed the activity title of _IdentitySettingsActivity_ from `SQRL - Settings` to `SQRL - Identity Settings` to reflect the above change
- Added an `IdentitySelector` to the IdentitySettingsActivity

Note:

I took the descriptive texts for the settings from Steve's windows client and adapted them to better reflect the situation on Android. If you have any suggestions for improvement, please let me know.


kalaspuffar:
Looked at the pull request and I can't see anything obvious to improve. Have not read the texts yet but will do that.

I feel that we are adding a lot of accessibility tech dept at the moment.
Guidelines say that clickable areas should be > 48dp in height and width. Or larger than 9mm to give everyone a chance of hitting the right controls. 

We also need to be careful with the use of contrasts so everyone can read text.

A tool you can use is the Accessibility Scanner available in the play store.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar , 

thanks for the input on the accessibility topic. This is something that was bugging me already with the identity selector. I knew that the touch targets there were too small, but did not want to sacrifice the compact form factor.

The Accessibility Scanner is a great suggestion, I will definitely give this a go and see what can be improved in that regard.

Best regards,
Alex

kalaspuffar:
Hi again @alexhauser 

First of please fix the merge conflict, probably just added strings from the other pull request.

Then can you look at the landscape layout so you did the same fixes in that design as you did in the portrait layout? 
I had to fix the ```layout-land/activity_url_login.xml``` as a selector was missing there so it crashed when I tested on my emulator.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

I've just resolved the merge conflict.
And this time I didn't even overlook the landscape layout ;-)

settings_landscape

alexhauser:
Hi again @kalaspuffar ,

I'd still like to push an accessibility update to this PR, please wait with merging this, I'll post again if we're ready to roll.

alexhauser:
Hi @kalaspuffar ,

accessibility is now handled as well (with the exception of `android:contentDescription`, which I left unhandled intentionally to avoid having our translators to have even more work).

PR should be ready for merging now.

Best regards, 
Alex

kalaspuffar:
Hi again @alexhauser 

I might be nitpicking now but I just looked at the result in the emulator for a tablet device and does this not look a bit strange. The selector might grow to 30 characters and take less space than something that might be maxing 5 characters.

And separating the QuickPass length and QuickPass timeout values seems a bit strange too. If you want to do something with this before the merge. I think there could be more done with this settings page so it might not be critical.

What do you think?

settings

alexhauser:
Hi @kalaspuffar ,

I don't mind any nitpicking at all, if it is benefitting the project. Especially since I am somewhat of a nit-picker myself ;-)

I agree that the landscape layout can be improved. I basically just blindly copied what was already there and slapped in the selector without too much thinking. But I already see some possibilities for improving the layout. 

You can expect another commit addressing the issue shortly.

alexhauser:
Hi @kalaspuffar ,

everything I've tried has pros and cons. 

Letting the selector use the whole width is certainly better, and I've also tried to make the grouping a bit more logical. Also, I added a ScrollView around the whole layout to at least make it workable (if not beautiful) on tiny resolutions.

Phone:

id_settings_phone_land

Tablet:

id_settings_tablet_land

If you have any further suggestions, please let me know!

sesam:
Since the new version is a great improvement, maybe it would be worth to merge it and try it out a bit, before we rework it completely. :).
Is it normal to have both Save and Cancel be the same color? On web apps, we would have the primary button filled with the primary color, and the secondary (Cancel) would instead be outlined with the primary color. That's the first thing I noticed - having to read and remember which button is which and and worry that I'll need to read it every time to hit the right one.

alexhauser:
Hi @sesam,

thanks for your input and good question regarding the button colors. I will look into this. I like the idea of having the secondary button only outlined.

Hi @kalaspuffar, I have just resolved the merge conflict regarding the new strings, should be ready to go again. I guess I'll make the button issue which Simon brought up a separate PR and address it everywhere where appropriate instead of in settings only.

kalaspuffar:
Hi @alexhauser 

Sounds like a plan.

Best regards
Daniel

PR #366, opened 2019-06-27
Fix build error (es language string)

alexhauser:
Description:

The latest merge broke the master build again.
This time, the culprit was a broken spanish translation.

Changes:

Fixed the broken language string by resetting it to the english version


kalaspuffar:
Great catch @alexhauser . This pull request will not help though as we generate strings from poeditor. I'll look at the translation and see if I can improve the import script.

kalaspuffar:
I've added a download to the script and more languages but this is the main script I'm using.

https://github.com/kalaspuffar/poeditor-translate

kalaspuffar:
I removed the input on POEditor. Someone started a translation but never completed it. So this fix will be valid even after I export.

Issue #363, opened 2019-06-26
Verify that Lock, Unlock and Remove account work


kalaspuffar:
As we have done a lot of changes these features have not been tested recently and might have stopped working. Got a report in the forum about Locking not working.

Will verify and fix any issues arising.

PR #370, opened 2019-06-28
Fixing ask question and lock, unlock and remove.

kalaspuffar:
Issue #363
Fixing so lock, unlock and other things that require the answer variables to be set in order to do the next call.

Description:
We reset the response values before we had actually used them, we can't reset any values before we have done the next call.

Changes:
Removed code that cleared response.
Added check for empty ask dialog.


kalaspuffar:

A look at this one would be nice. 

Good to get these features working again :)

One text string for missing account is not added in this pull request will add that by downloading translations from POEditor add push that later today.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

I still feel not quite competent enough to comment on any of the communication code, I just haven't dug into it deep enough yet.

But what i can and did do is test, and this is what I've found:
All the functions (lock, unlock and remove) are basically working. However, there are still some quirks which should be fixed:

- After disabling my account on GRC's demo site, I was redirected back to the website, which apparently just reloaded the page, but neither our app nor the website gave me any indication that the account was successfully locked. I don't know if the protocol allows it, but it would be nice for the app to show a confirmation message upon successfuly locking the account before finishing and going back to the website. The same is true for unlocking the account of course.

- After unlocking the account, I tried to log in again, but after going through the login in the app, it just directed me back to the website, which - again - apparently just reloaded the login page but never logged me in. To fix this, I had to start the app from the launch icon (since the login activity does not appear in the recent apps anymore), and then clear it from memory by swiping it away in the recent apps. Only after that, the login would start working again.

- Lastly, we have some inconsistent terminology in our app: On the action buttons it says "lock" and "unlock", while on the activity titles it's "disable" and "enable" etc.

Hope this helps.

Best regards,
Alex

Issue #369, opened 2019-06-28
Handle scanning an identity qr-code from the main screen


alexhauser:
### Feature request.

See https://sqrl.grc.com/threads/import-via-qr-fail.716/post-5957

> Is it possible, when the QRCode decodes to something without a nut= embedded, that there could be an error message that explains something (roughly) like: "This does not appear to be a SQRL authentication QRCode. If you are attempting to import your identity, please use the QRCode scanner function in the Import Identity process."

I would even like to try to hand off the job to the import activity as soon as scanning a qr code by the main "scan qr code for site" button determines that an identity qr code has been sacnned.

So in the end, you wouldn't even have to explain anything, and the app would just do the right thing for the type of barcode you're scanning.

PR #375, opened 2019-06-29
Handle scanning an identity qr-code from the main screen

alexhauser:
Issue #369.

Description:

Handle scanning an identity qr-code on the main screen scan button by "forwarding" the qr-code data to the `ImportActivity`.

Changes:
- Add new import method named `IMPORT_METHOD_FORWARDED_QR_CODE` as intent extra for `ImportActivity`; Also, I've moved all the static strings used for intent extras from `ImportOptionsActivity` to `ImportActivity`
- Detect the type of qr-code in `SimplifiedActivity` and forward data to `ImportActivity` if an identity qr-code was detected, adding the above import method extra to the intent
- Handle a forwarded identity qr-code in `ImportActivity` if present in the intent extras

I also noticed that we were still using the `readSQRLQRCodeAsString(Intent data)`, which did not iterate over the qr-code byte segments, but instead would only `return data.getStringExtra("SCAN_RESULT")` and therefore sometimes delivered truncated data. I have changed the implementation of this function to use the `readSQRLQRCode(Intent data)` under the hood, which was recently updated to deliver correct results under all cirumstances.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/SimplifiedActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/SimplifiedActivity.java
@@ -7,14 +7,15 @@ import android.os.Bundle;
 import android.os.CancellationSignal;
 import android.support.design.widget.Snackbar;
 import android.util.Log;
-import android.widget.ImageButton;
 import android.widget.TextView;
 
+import com.google.zxing.FormatException;
 import com.google.zxing.integration.android.IntentIntegrator;
 import com.google.zxing.integration.android.IntentResult;
 
 import org.ea.sqrl.R;
 import org.ea.sqrl.activites.base.LoginBaseActivity;
+import org.ea.sqrl.activites.identity.ImportActivity;
 import org.ea.sqrl.processors.BioAuthenticationCallback;
 import org.ea.sqrl.processors.CommunicationFlowHandler;
 import org.ea.sqrl.processors.CommunicationHandler;
@@ -24,6 +25,7 @@ import org.ea.sqrl.utils.SqrlApplication;
 import org.ea.sqrl.utils.Utils;
 
 import java.security.KeyStore;
+import java.util.Arrays;
 import java.util.regex.Matcher;
 
 import javax.crypto.Cipher;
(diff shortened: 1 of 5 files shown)


kalaspuffar:
Hi @alexhauser 

I've reviewed the code and the function readSQRLQRCodeAsString reads a URL and can be cut-off if we use the pure data stream. I had some problems with that when implementing against WordPress plugin. 

So when reading strings / URLs we use the pure text stream that XZing gives us. We have no need to fetch the raw data in order to get this information. Using the byte stream is required to get the identity code though. I had a hack earlier to retrieve QRCode data through the string interface but that's not the right way.

So I suggest you revert that change and I'll do some testing on my end. Or have you found a problem with using the string interface for the URL fetching?

Best regards
Daniel

alexhauser:
Hi @kalaspuffar , 

thanks for reviewing. The reason why I changed `readSQRLQRCodeAsString(Intent data)` was that it was truncating the identity qr-code data. Only the recently fixed version of `readSQRLQRCode(Intent data)`, which uses the raw byte segments, was giving me the desired result for both login- and identity qr codes during testing.

So, since `getStringExtra("SCAN_RESULT")` was truncating, I felt it shouldn't be used at all, and we should only ever use the "safe", non-truncating raw bytes segments and simply do the String convertion ourselves in the helper method. 

The truncating of raw bytes that you're talking about was fixed with one of your latest commits, by looping over all available data segments, wasn't it? Or did you also experience cutoffs using the fixed version of  `readSQRLQRCode(Intent data)`?

Best regards,
Alex

kalaspuffar:
Hi again @alexhauser 

I think I need to test it again, I had the opposite experience but I'll test it again.

Which URLs did get truncated in your tests? Have you tried with the URLs at https://uhash.com?

I'll do my own tests as well. been thinking a lot about work today but I'll do some tests during lunch now.

Best regards
Daniel

kalaspuffar:
Hi @alexhauser 

Did some testing and it seems to work just fine. So might be a problem on my side. The first time I scanned a QR identity the application crashed on one of our memory constrained devices. But after that this PR worked just fine.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

thanks for testing and confirming that it works. I also did a quick test on  https://uhash.com using the code in this PR and it logged me in just fine. 

Do you think this PR is safe to merge, or should I do some investigation on https://github.com/journeyapps/zxing-android-embedded first to confirm our conclusions?

Best regards,
Alex

kalaspuffar:
I say we merge, if we don't see any issues it should be an improvement. If you want you can look at the code, it's good that we have knowledge about the best usage for QR Code scanning. If we find a better way we can always create another pull request.

Issue #338, opened 2019-06-06
Consolidate login UI and code


alexhauser:
This is somewhat related to issue #333 , but should probably be done before anyone starts working on #333, only to make things easier for them.

*Description* 

We currently support 2 ways of logging in:
- By scanning a QR code displayed by another device (CPS = false)
- By clicking on a sqrl link on the Anrdoid device itself (CPS = true)

We have separate UI and logic code for both of those login methods, despite the fact that they are largely identical.

*Suggested enhancement*
Abandon the login popup window in `SimplifiedActivity` and use the `UrlLoginActivity` instead also for scanning a QR code. This could be done easily by just starting an intent to launch `UrlLoginActivity` after having received the sqrl:// link from the QR code scan:

``` java
    Intent urlIntent = new Intent(Intent.ACTION_VIEW);
    urlIntent.setData(Uri.parse(serverData));
    urlIntent.putExtra("EXTRA_USE_CPS", false);
    startActivity(urlIntent);
```
The extra which is put in the intent can be detected in `UrlLoginActivity` and CPS could be turned off accodingly.

This would get rid of a lot of redundancy and leave us with only one login codebase to deal with.

What do you guys think?

sengsational:
This is a great idea!  I had the same thoughts as I was working on the shortcuts, but powered through it without making this set of changes.

kalaspuffar:
Hi @alexhauser and @sengsational 

Well, this is another point of the legacy that will be cleaned up by a better approach. I say go ahead.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar, you can assign me to this issue, I've already started working on this. It could take a few days though, because I'm on holidays with my family and away from my dev tools.

sengsational:
I hope you enjoy your holidays Alex!  I'm going to be away quite a bit of time myself, coming up 17 June through 2 July.  I'll be home some of that time, but mostly away.  And Daniel, I didn't mention it in the other thread, but congratulations on the new nurse!  That's a laudable profession.  My mom and nephew are nurses.

alexhauser:
Thanks for the heads-up Dale, and have some wonderful holidays yourself, too, once you get going!

Daniel, congrats to your wife also from Austria! 😊

alexhauser:
Hi @sengsational and @kalaspuffar , 

[...]

PR #378, opened 2019-07-07
Remove logon popup

sengsational:
Issue #338.
**Description:**

Continues building on Alex's 9ae1fef migration away from login popup to the `UrlLoginActivity`.

**Changes:**

* Removed linkages from shortcut code to login popup, using `UrlLoginActivity` instead.
* Removed unused login popup code (removed `setupLoginPopupWindow()`)

**Discussion:**

I had to remove `setupLoginPopupWindow()` from `AlternativeLoginActivity.onCreate()`.  I'm not sure when and how that code is called, so didn't test that.  Removal of `setupLoginPopupWindow()` from `SimplifiedActivity.onCreate()` tested out successfully.


app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
```diff
@@ -132,36 +130,19 @@ public void afterTextChanged(Editable s) {
             long currentId = SqrlApplication.getCurrentId(this.getApplication());
 
             if(currentId != 0) {
-                doLogin(storage, txtLoginPassword, false, true, null,this);
+                doLogin(storage, txtLoginPassword, false, useCps, true, null,this);
             }
         });
 
-        if(Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && storage.hasBiometric()) {
+        if(Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && storage.hasBiometric() && !ACTION_LOGON.equals(intent.getAction())) {
```
alexhauser: I think we should move the whole biometric stuff into the first `if (!ACTION_LOGON.equals(intent.getAction())) { 
}` block above. This way, we don't need a second test for the action, and we also eliminate the variable scoping issues. So you could leave stuff like `final String domain = sqrlMatcher.group(1);` in and actually use it in the biometric block, instead of having to fall back on the nondescript `mSqrlMatcher.group(1)`. What do you think?

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
```diff
@@ -132,36 +130,19 @@ public void afterTextChanged(Editable s) {
...
-                        communicationFlowHandler.handleNextAction();
-                    });
+                    new BioAuthenticationCallback(UrlLoginActivity.this.getApplicationContext(), () ->
+                        doLogin(storage, txtLoginPassword, false, useCps, false, UrlLoginActivity.this, UrlLoginActivity.this)
+                    );
 
             BiometricPrompt bioPrompt = new BiometricPrompt.Builder(this)
                     .setTitle(getString(R.string.login_title))
-                    .setSubtitle(domain)
+                    .setSubtitle(mSqrlMatcher.group(1))
```
alexhauser: See comment above.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
```diff
@@ -97,14 +103,6 @@ protected void onCreate(Bundle savedInstanceState) {
             txtLoginPassword.setHint(R.string.login_identity_password);
         }
 
-        txtLoginPassword.setOnEditorActionListener((v, actionId, event) -> {
```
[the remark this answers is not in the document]
sengsational: My mistake.  You are correct.  Needs to get put back.  I even noticed this when testing but didn't make the connection.  Thank goodness we've got more eyes on these changes than just mine.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
```diff
@@ -56,33 +59,36 @@ protected void onCreate(Bundle savedInstanceState) {
             return;
         }
 
-        Uri data = intent.getData();
-        if(data == null) {
-            showErrorMessage(R.string.url_login_missing_url);
-            return;
-        }
-        txtUrlLogin.setText(data.getHost());
+        if (!ACTION_LOGON.equals(intent.getAction())) {
```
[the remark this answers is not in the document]
sengsational: Great point.  I think it should be changed per your suggestion.  It's certainly confusing as-is.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
sengsational: Rather than struggle with this, I just "made it work", which, of course, isn't the right way to proceed.  If I had seen the opportunity you point out of moving all the biometric stuff in to the first block, I would have done it, but the idea didn't come to me. 

I set this PR to "allow collaborators to change" or something like that.  Would you / could you make this change @alexhauser?

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
sengsational: Yes, I was thinking of adding a `getDomain()` method to the matcher to get around this, but I like the idea of restoring the `String` definition and moving all the biometric code.

Just a wild, but maybe related idea... you know those stack traces thrown because biometric code is inside an activity?  If you do go about moving the biometric stuff, you might consider looking at the _concept_ used in `LoginActivityHelper` (https://github.com/sengsational/secure-quick-reliable-login/tree/remove_starting_stacktraces) to prevent those stack traces.  That code was never merged, because it's not really "our fault" that these stack traces are thrown.  Still, if it makes sense, pulling the biometrics out of the activity itself might be an option.

sengsational:
True, @alexhauser , there are some unused traces of the logon popup remaining.  If this, so far, looks reasonable (given the changes you've suggested), cleanup of those things can be commenced.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
```diff
@@ -30,15 +28,20 @@
 
 import javax.crypto.Cipher;
 
+import static org.ea.sqrl.activites.SimplifiedActivity.ACTION_LOGON;
```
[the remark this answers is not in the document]
sengsational: Yep, it's in the wrong place.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
```diff
@@ -231,4 +204,19 @@ protected void onStop() {
             hideProgressPopup();
         }
     }
+
+    private IdentitySelector configureIdentitySelector(SQRLStorage storage) {
```
sengsational: One thing I forgot to mention in the original message in this thread was that I came across an instance when resuming `UrlLoginActivity` where `mIdentitySelector` was null.  That precipitated `configureIdentitySelector()`.  I was considering just not having an instance variable at all, rather just building it as needed.  Considered (briefly) the idea of changing the `IdentitySelector` to a singleton pattern, but wasn't sure if that was appropriate (didn't think real hard about it).

alexhauser: I am completely fine with what you did with the identity selector, I think the `configureIdentitySelector()` solution is appropriate to avoid ending up with an uninitialized selector, which I guess could have happend if onStop() was called upon leaving the activity without the system killing the app process and then the user navigates back to the activity.

I'm glad you caught this and will be more careful with this stuff in the future.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
alexhauser: I guess I'll have to clone your forked repo and then I should be able to push to your feature branch. I'll have a look asap.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
alexhauser: Good point, Dale. If I remember correctly, the PR was never merged because there was some other controversial stuff in it. The biometric stuff everyone was fine with as far as I remember. 

Me personally, I would appreciate pulling the biometrics stuff out into its own helper class, but would then call it something like `BiometricLoginHelper` to better reflect what it acutally does.

However, I think we should not be adding more stuff into this PR at this time, but instead do all of these things as small follow-up PRs that are not tangled up with the consolidation stuff which is at the center of this PR, and will therefore be much easier to review and test.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
alexhauser: Thinking about this again, I guess I'll leave this as-is for now and post a follow-up PR with the mentioned optimization later instead of squeezing it in here. Small diffs ftw.

alexhauser:
I've just tested the latest version of this PR and found 

- URL login using password
- URL login using QuickPass
- URL login using biometrics
- URL login using Alt-ID and password

as well as

- QR-code login using password
- QR-code login using QuickPass
- QR-code login using biometrics
- QR-code login using Alt-ID and password

all to be working as expected.

146 lines added, 655 lines removed - not too shabby! :)

For me, this is ready for merging.
(Some optimizations will come later as small follow-up PRs)

kalaspuffar:

Sadly this PR is not working. I guess the login feature is OK but sadly the other functions are not working.

We switched over to using only the on-device login flows. This means that disabling, enabling and removing accounts give you CPS errors even if you scan QR codes. Another thing that is less desirable is that after each of these functions the application closes, which is correct if you go back to a page but not if you use the application without CPS.

Lastly, the noiptest flag is not set for these calls which makes them fail.

Great work so far but we need to give this a bit more work.

I'll wait to review the code before we can get all flows working.

Best regards
Daniel

kalaspuffar:

I think the issue is that we load the SimplifiedActivity and run the code ```setupBasePopups(getLayoutInflater(), false);``` later when we open the UrlLoginActivity we run ```setupBasePopups(getLayoutInflater(), true);``` this negates the feature so we will actually not remember our preference.

Best regards
Daniel

alexhauser:
Great catch, @kalaspuffar !
@sengsational, is it possible for you to fix this?

sengsational:
I'm not confident that this one going to have a clear solution  that comes to *my* mind; I don't have a complete understanding of how the original solution worked.

The basic idea, as I understand it, is that you can scan a QR code -or- you can get a message from a browser.  So the Boolean `urlBasedLogin` is "true" for the latter, and "false" for the former?

If the above is correct, then it seems like we're configuring the parameters of `CommunicationFlowHander` (through `setupBasePopups()`) "too early" in the process...before we know if it's a QR code or a web browser sourced login.  

Sorry for being amorphous with my descriptions...this is something I have not taken the time to completely understand.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
sengsational: Agreed.

alexhauser:
> If the above is correct, then it seems like we're configuring the parameters of CommunicationFlowHander (through setupBasePopups()) "too early" in the process...before we know if it's a QR code or a web browser sourced login.

I think that's exactly what's happening. However, I don't even think that we're configuring it too early, we're just passing a hardcoded `true` to the `urlBasedLogin` parameter, even though we already have determined the login type at that point, so I guess we should just be able to pass the `useCps` boolean instead of the hardcoded value in `UrlLoginActivity`. I will do some testing and report back once I'm confident that this will be sufficient to fix the issue.

alexhauser:
Hi @sengsational and @kalaspuffar ,

my testing showed that ideed changing `setupBasePopups(getLayoutInflater(), true);` to `setupBasePopups(getLayoutInflater(), useCps);` in BOTH `onCreate()` and `onResume()` does correct the issue with lock, unlock and remove.

Also, adding the following code to the very top of the existing `closeActivity()` solves the problem of the whole app closing after a qr-code login:
```java
    if (!useCps) {
        super.closeActivity();
        return;
    }
```

The whole `closeActivity()` function should then look like this:
```java
    @Override
    protected void closeActivity() {
        if (!useCps) {
            super.closeActivity();
            return;
        }

        UrlLoginActivity.this.finishAffinity();
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
            UrlLoginActivity.this.finishAndRemoveTask();
        }
    }
```

Dale (@sengsational), can you please put in these small fixes, saving me from needing to clone your whole repo just to push a small commit to the PR's branch? Thanks!

Best regards,
Alex

sengsational:
Alex, Thanks for getting that analysis and solution done.  I'll make the changes now.

sengsational:
Those few changes (with slight modifications) were made in 27986ae.  I liked how when doing a scan from the shortcut the application closed (both started and finished right on the Android home screen), so I added a bit of code to make it so if the user used CPS or used the application level shortcut to scan, the application "went away".  But if the user scanned from within the application, the application does not disappear, which I agree, was something a user would probably not expect.

I came across an idea for improvement on `closeActivity()`, but I resisted the temptation to make this PR even bigger.  I'll need to do another change before this goes in because of the comment I left in :(

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
```diff
@@ -176,15 +177,22 @@ public void afterTextChanged(Editable s) {
...
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-            UrlLoginActivity.this.finishAndRemoveTask();
+        boolean quickScan = getIntent().getBooleanExtra(EXTRA_QUICK_SCAN, false);
+        if (useCps || quickScan) {
+            UrlLoginActivity.this.finishAffinity();
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+                UrlLoginActivity.this.finishAndRemoveTask();
+            }
+        } else {
+            Log.v("sengsational", " returning 'early'");
```
sengsational: Oops...I'll need to remove the logging entry before this gets merged!

kalaspuffar:
I'm reviewing now so please tell me @sengsational when you are done so I might merge it.

sengsational:
I'm done, except for there's a `Log.v()` that needs to be removed.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
```diff
@@ -56,36 +59,39 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        communicationFlowHandler.setServerData(serverData);
-        communicationFlowHandler.setUseSSL(serverData.startsWith("sqrl://"));
+            useCps = intent.getBooleanExtra(EXTRA_USE_CPS, true);
 
-        Matcher sqrlMatcher = CommunicationHandler.sqrlPattern.matcher(serverData);
-        if(!sqrlMatcher.matches()) {
-            showErrorMessage(R.string.scan_incorrect);
-            return;
-        }
+            txtUrlLogin.setText(data.getHost());
```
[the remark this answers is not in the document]
sengsational: I'll keep an eye out for changes with this.  Thanks for the heads-up.

app/src/main/java/org/ea/sqrl/activites/UrlLoginActivity.java
```diff
@@ -56,36 +59,39 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-        final String domain = sqrlMatcher.group(1);
-        final String queryLink = sqrlMatcher.group(2);
+            final String serverData = data.toString();
+            communicationFlowHandler.setServerData(serverData);
+            communicationFlowHandler.setUseSSL(serverData.startsWith("sqrl://"));
```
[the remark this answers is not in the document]
sengsational: I'll research this a bit; it's not something I have analyzed.

sengsational:
Daniel, Thanks for reviewing and merging!

PR #381, opened 2019-07-09
Fixing rekey function.

kalaspuffar:
Description:
When you rekey an identity this identity will have an old previous key and then the new key is put into our secret vault. This means that you will have new suk (server unlock key) and vuk (verify unlock key). These must be communicated with the server when a previous key has been found to be valid.

Changes:
Set a boolean in Storage when a previous key has been found.
Send suk and vuk when previous key is sent to the server.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/processors/CommunicationFlowHandler.java
+++ b/app/src/main/java/org/ea/sqrl/processors/CommunicationFlowHandler.java
@@ -256,6 +256,8 @@ public class CommunicationFlowHandler {
             actionStack.push(a);
             lastTIF = commHandler.getTif();
             commHandler.clearLastResponse();
+        } else if(commHandler.isPreviousKeyValid()) {
+            commHandler.loginWithPreviousKey();
         }
 
         if(commHandler.hasAskQuestion() && this.actionStack.isEmpty()) {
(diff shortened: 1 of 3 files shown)


kalaspuffar:

This code might be a bit low level but if you want and can find anything obvious that I have missed or forgot it would be appreciated if you take a look. Would like to get these merged and released this week.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar,

I'm still making myself familiar with the low level code within the project, so a review will be superficial at best.

But I will try to test the PR tomorrow!

Best regards,
Alex

sengsational:
I have not looked much at this portion of the code either, nor have I done any testing associated with re-keying, although I might use this as an opportunity to learn more about it.

alexhauser:
Hi @kalaspuffar ,

I was able to successfully rekey two test identities, so this PR looks good in terms of operability.

Here are some things that came to mind during the rekeying process:

- As it is now, we are not replacing the identity with the new, rekeyed identitiy but rather adding a rekeyed identity and leaving the old one there. Is this intentional? I would have expected the rekeyed identity to replace the old one. 
But on the other hand, this way we're on the save side and have all the previous data should something go south.

- Should we decide to replace "old" identity, we should hide the "Identity name" field upon saving the new identity and use the "old" identity's name.

Best regards,
Alex

kalaspuffar:
Hi @alexhauser 

Keeping the old one was intentional for safety. This is a very rarely used feature when you think you've been compromised. So keeping your old one should not be an issue.

Easy thing to change if we like.

Best regards
Daniel

Issue #379, opened 2019-07-07
Alt-ID is not using QuickPass


alexhauser:
### Feature request.

Currently, when logging in to a site using the Alt-ID feature, we always ask for the full password, even if QuickPass is active. I guess we should fix this.

kalaspuffar:
Hi @alexhauser 

Yes we can. I think the more pressing matter though is getting the disable, enable and remove to use the ALT-ID account.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

you're absolutely right, this didn't even occur to me.
But I think I might have a good idea for how we could elegantly solve all of this. I'll come up with a suggestion.

Best regards,
Alex

alexhauser:
Here is my suggestion for solving the alt-id  issues discussed above:

In the login screen, we remove the "options" button and add an (initially collapsed) area called "advanced functions" or similar:

<img src="https://user-images.githubusercontent.com/4005543/60894109-5e0e4900-a262-11e9-9e8a-14c98057a3b7.JPG" width="300">

On expanding the advanced funtions, we can enter an alt-id and also choose between the different account options (and get help using the info icons):

<img src="https://user-images.githubusercontent.com/4005543/60894155-74b4a000-a262-11e9-905f-e223a65cd7e9.JPG" width="300">

For the account options that need the resuce code, we just show an "enter rescue code" popup once the user clicks the login button (which will be dynamically labelled according to the chosen function of course).

This would provide everything in one nice login flow and enables providing alt-id functionality for account options, all without needing to jump into different activities. 

And, we can get rid of 5 activities using this approach!

@kalaspuffar , @sengsational and @sesam ,
what do you guys think?

sesam:
Looks user friendly! How does the Altwrnative identity oart work? Free text input? Some autocomplete? I'm basically a tiny bit confused around the thin line with gray text Alternative identity.

alexhauser:
Hi @sesam, it's just a standard `TextEdit` field for free text input, since the alternative id can be anything you like.

kalaspuffar:
Hi @alexhauser 

Concept, awesome. The design which might not be my expertize but the gray feel a bit "heavy" so choosing a lighter shade could be good.

In the dark design, a light border would do the same effect.

Best regards
Daniel

sengsational:
Looks like some very good thinking there, Alex; I like it.

[...]

Issue #371, opened 2019-06-28
Disabling account creates bad state


kalaspuffar:
### Issue or error solving.

*Short description of the issue*

After unlocking the account, I tried to log in again, but after going through the login in the app, it just directed me back to the website, which - again - apparently just reloaded the login page but never logged me in. To fix this, I had to start the app from the launch icon (since the login activity does not appear in the recent apps anymore), and then clear it from memory by swiping it away in the recent apps. Only after that, the login would start working again.

*Expected behavior*

Was able to log in

*Actual behavior*

Could not log in until app was reset.

*Steps to reproduce*

1. Disable account
2. Enable account
2. Try to login again

kalaspuffar:
QuickPass issue?

alexhauser:
Could be. It always happend with QuickPass engaged.

alexhauser:
Hi @kalaspuffar ,

this is still an issue even after PR #382. I also don't think it's QuickPass-realated any more, since all my mitigation in that direction were fruitless. It's probably something rather low level (storage or comm), could you please look into this sometime when you're back from your vacation and find some time to spare?

Best regards,
Alex

kalaspuffar:
Hi @alexhauser 

I tried this issue with the new version and could not recreate it so it seems the new flow solved my problems. I'll close this.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

I've just verified that it indeed works as expected now! Thanks!

Best regards,
Alex

mozzworld:
@kalaspuffar, @alexhauser,
After the diable, enable sequence, I had the same experience (could not login) until I cleared the browser cache. I was using version 1.5.1 downloaded from F-Droid to login to sqrl.grc.com - @mozz sqrl.grc.com

kalaspuffar:
Hi @mozzworld 

Seems to be an issue with the website if a clearing of browser cache can solve the issue.

Best regards
Daniel

mozzworld:
Hi. @kalaspuffar,

Now that I know how to work around, it's okay. As things are still in progress, I am hopeful such issues will be taken care of in due course.

With regards,

Mozz

PR #382, opened 2019-07-14
Better alt-id and account options

alexhauser:
Issue #379.

This one looks a little scarier than it actually is, mainly due to some refactorings that I've done along the way. Still, sorry for the rather large PR, but I hope you'll agree that it's worth it in the end.

**Description:**

Move alternate-id as well as account options (disable, enable and remove) to the main login flow. 

This solves the issue that none of the account options could previously be used together with an alt-id. Also, it enables using QuickPass for any operation that doesn't require the RescueCode. 

Lastly, this should improve user-friendliness as well as maintainability, since we now have all the login-related code in one place.

**Changes:**

- Add an (initially collapsed) area called "advanced functions" to the login window
This contains the text field for the alt-id as well as all the account-related options
- Add an (initially hidden) rescue code input form to the login window
This will be shown instead of the password field on operations that need the RescueCode
- Add help icons to all advanced functions, providing user guidance
- Rename `UrlLoginActivity` to `LoginActivity`
- Remove the now obsolete activities 
    - `AlternativeLoginActivity`
    - `AccountOptionsActivity`
    - `DisableAccountActivity`
    - `EnableAccountActivity`
    - `RemoveAccountActivity`
- Adapt `EnableQuickPassActivity` to the new layout of `LoginActivity`

When testing this, please note  that issue #371 is still open, so the app needs to be reset after disabling an account, otherwise all the following actions will fail.

Here are some screenshots of the new login window, when advanced functions are expanded:

Light mode, operation requiring only the password:

Light mode, operation requiring the RescueCode:

Dark mode, operation requiring only the password:

Dark mode, operation requiring the RescueCode:


kalaspuffar:
Hi @alexhauser 

I ran the PR now when I'm back at the keyboard. And found the following strang background on the identity name. 

And if you click on advanced options the description text is removed. Perhaps that is intentional but it as a user it is something you notice and wonder why. The text will not return if you press the advanced options again and close that menu.

Will look at the PR code more closely this evening.

Best regards
Daniel

background-issue

kalaspuffar:
decription-shown

alexhauser:
Hi @kalaspuffar,

thanks for running the PR and commenting.

The background behind the identity name is weird, I haven't seen this in my testing, but I will investigate.

Hiding the description (and not showing it again) was intentional, but can be discussed of course. I figured anyone using advanced features wouldn't need the description anyway, and we can use the screen real estate. But I'm completely open for changes here!

Best regards,
Alex

alexhauser:
By the way, I cheated a little on the accessibility front on this one, since making every radio button 48dp high would have bloated the layout a lot. But this is also something that can be discussed of course.

alexhauser:
Hi @kalaspuffar ,

I just wanted to let you know that while testing, I "accidently" found out what the gray background that you had experienced on some controls a while back is actually all about: It's simply **input focus**. 

When we are testing with the emulator and using the physical keyboard, Android leaves "touch mode" and enables input focus on controls. I noticed by chance that, when switching through controls with the tab key on the physical keyboard, some touch-enabled `TextView`s would get input focus, which is indicated by displaying a light gray background on the control. 

So this is in fact "intended behaviour" and we can erase this mistery from the back of our heads. 😉

Non the less, I will set a few `android:focusable="false"` attributes where input focus is not needed (as for example on the identity name in the identity selector, since we still have the icon which can be focussed in non-touch-mode, and this does not affect that it can be tapped/clicked).

Best regards,
Alex

kalaspuffar:
Hi @alexhauser 

Awesome catch. Great work, it was something that annoyed me even though it wasn't important but having a reason feels much better.

Thanks
Daniel

Issue #380, opened 2019-07-09
Scanning identity from a darkmode device


kalaspuffar:
### Issue or error solving.

*Short description of the issue*
When the phone is in darkmode the QR Code will show white on dark background which work less well with some devices.

*Expected behavior*
Scanning a QR code should work on an older device when presented with a QR Code from a new device in darkmode.

*Actual behavior*
Scanning fails.

*Steps to reproduce*
1. Set phone 1 in dark mode.
2. Try to scan QR code from another device.

kalaspuffar:
A solution could be to set a thin border white around the QR code. That might work.

sengsational:
How do you set the phone to dark mode?  My phone always seems have the same color scheme, no matter what time of day it is.  I suppose I could dig into the code, but in case it's something really easy that I'm overlooking, I thought I'd ask.

kalaspuffar:
Hi @sengsational 

Some phones can have dark mode during the evening.

With the new Q release you can turn dark mode on for all uses and apps. I don't know if all have this beta installed. If not then I can look into it :)

Best regards
Daniel

sengsational:
I did a bit of work on this.  The basic problem was just a couple of lines in the `activity_show_identity.xml`, but in order to test it on my phone, I had to figure out a way to get my phone running in night mode.  I added a method that runs on startup to put the phone into automatic mode (my phone isn't in that by default).  I won't know until sundown if that works :-)  But the way I was able to test was by adding a thing that toggles the night mode; by tapping repeatedly/quickly on the app, it will toggle night mode.  

There are hard-coded strings in the method, but I didn't know if it would be advisable to leave the app-tester-only code in the app, or remove it, so didn't want to bother with adding those to all of the language string files.

No PR at this time because there's more to be discussed, but I did push the code to my repo: 
https://github.com/sengsational/secure-quick-reliable-login/tree/dark_mode_scanning

kalaspuffar:
Hi @sengsational 

```
AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_YES);
```

Maybe forcing the mode during development could be a thing?

Best regards
Daniel

sengsational:
I tried that one, but `UiModeManager` seemed to work better for me.

alexhauser:
Hi @sengsational , 

[...]

PR #391, opened 2019-07-20
display QR code with white background

sengsational:
Issue #380 

**Description:**
Add white background around QR code for night-mode scanning.

**Changes:**
Added background to portrait and landscape QR code layouts.


alexhauser:
I've tested the changes already in your repo and can confirm that they work as is.

Cheers, Alex

Issue #388, opened 2019-07-19
Properly handle cancelling a CPS authentication


alexhauser:
### Enhancement.

Steve describes on page 9 of the new `SQRL Operating Details` document how to properly handle an aborted CPS authentication:

> **What if no cancel parameter?**
... 
In this case, the SQRL client should return a simple webpage to the browser to show its users something like: “You have cancelled the SQRL sign-in. Please press your web browser’s BACK button to return to the previous login page. That simple page might also contain JavaScript to pop the browser’s page history stack to automatically return the user to the previous page.
... 

We currently redirect the browser to `www.google.com`.

kalaspuffar:
Hi @alexhauser 

Great catch. I've not read the text yet as it's still in it's final revisions and I just returned from my vacation :)

I don't know what page we should redirect to so I choose Google back in the day as it probably would be present. Creating a simple html page with information could be an easy fix though :)

Best regards
Daniel

alexhauser:
> Creating a simple html page with information could be an easy fix though :)

Yep, I'm on it!

alexhauser:
How about this, @kalaspuffar ?

![auth_aborted](https://user-images.githubusercontent.com/4005543/61576026-6e30ee80-aad4-11e9-8f49-d4f52d9df9b0.jpg)

kalaspuffar:
Hi @alexhauser 

The read in the title might be a bit deep. Then again the background is also deep. So maybe a lighter color would lighten up the page a bit?

Best regards
Daniel

alexhauser:
Hi @kalaspuffar, 

I wanted to stick with the SQRL logo's "signature" colors. That's why I chose that blue and the white. But I agree on the red being a bit deep. I'll experiment a bit with the title color and get back with a suggestion.

Best regards,
Alex

alexhauser:
Hi @kalaspuffar ,

I think this should do it:

![auth_aborted2](https://user-images.githubusercontent.com/4005543/61587728-9d586600-ab8f-11e9-98ca-695643f12201.jpg)

Issue #387, opened 2019-07-19
Drop CPS queries which conatain "Origin:" header


alexhauser:
### Issue or error solving.

Skimming through Steve's *"SQRL Operation Details"* document, I stumbled upon the following sentence regarding CPS authentication:

> ... Therefore, all SQRL clients **must** drop **any** query received which contains an “Origin:” header. 

I thought I would check if we follow that spec but could not find any hint of this within `CommunicationFlowHandler.startCPSServer()`.

Am I looking in the wrong place or does this need fixing? 

Btw, the whole paragraph on this issue (page 5) reads:

> As we noted above, when CPS mode authentication completes, the SQRL web server returns an HTTP 302 Found reply. Since this reply contains the authentication token, we must assure that this is only received by the web browser and not by any JavaScript that might be running on an
untrusted and possibly fraudulent sign-in page. If JavaScript were to use an XML HTTP Request (XHR), or similar, to query the SQRL web server at localhost:25519, it would obtain the HTTP 302
redirect information returned by the web server, which it could abuse. Fortunately, web browser
security strongly enforces the differentiation of any and all script-based queries from browser page fetch queries. ALL script-driven queries contain an “Origin:” header which is not under the script’s control – and browser page-fetch queries do not. Therefore, all SQRL clients must drop any query received which contains an “Origin:” header.

alexhauser:
Hi @kalaspuffar, 

I guess this one should have the "bug" label instead of "enhancement", since it can be used to steal the auth token/url using scripting.

Best regards,
Alex

PR #392, opened 2019-07-22
CPS fixes/enhancements

alexhauser:
Issue #387, #388.

**Description:**

Properly handle cancelling a CPS authentication and drop CPS queries which conatain an "Origin:" header.

**Changes:**
- Decouple the CPS server functionality from `CommunicationFlowHandler` by breaking it out into its own class, `CPSServer`.
- Add a static html page to the projects assets, which gets sent if CPS auth is aborted but no cancellation url is provided using a "can" parameter.
- Parse the headers in `CPSServer` and abort if an "Origin:" header is found


kalaspuffar:
Hi @alexhauser 

Great work. Have you tested the CPS functionality extensively?

I've not started to look at this pull request yet. But the master branch seems to work badly with my WordPress plugin.

Redirects to www.google.com every time. Might look closer on this later tonight.

Could be the Plugin code though so just asking :)

Best regards
Daniel

kalaspuffar:
Can be QuickPass related.

alexhauser:
Hi @kalaspuffar ,

> Great work. Have you tested the CPS functionality extensively?

Thanks! I would have loved to text extensively but I couldn't figure out how to do it without controlling server behaviour. What I have done instead was to manually control the booleans `mCancelCPS` and `params.containsKey("can")` just so that I could at least trigger all the code paths and verify that they work as intended. Those tests all went well. What I don't know is if the right circumstances will actually trigger the right code path "in the wild", that would need some control over the server afaiks.

And while I am getting more familiar with the codebase every day, I still don't fully understand all the comm logic interactions. I wanted to ask you anyway to take a closer look at all of this, also considering issue #371 is still lurking.

And since you now have a server implementation as well, who would be the better man for the job? 😁

If i can help in any possible way, please let me know!

Best regards,
Alex

alexhauser:
Hi @kalaspuffar ,

> But the master branch seems to work badly with my WordPress plugin.
Redirects to www.google.com every time. Might look closer on this later tonight.

Hmmm strange, did not see this behaviour in any of my testing. But I did only test against the GRC demos and the forums, not against your wordpress plugin. 

Looking forward to hearing back after you've looked at it!

Best regards,
Alex

app/src/main/java/org/ea/sqrl/processors/CPSServer.java
```diff
@@ -0,0 +1,257 @@
...
+                mServerSocket = new ServerSocket(CPS_SERVER_PORT);
+
+                while (!mServerSocket.isClosed() && !done) {
+                    Socket socket = mServerSocket.accept();
+                    BufferedReader in = new BufferedReader(new InputStreamReader(socket.getInputStream()));
+
+                    String requestLine = in.readLine();
+                    Log.i(TAG, requestLine);
+
+                    Map<String, String> headers = parseHeaders(in);
```
kalaspuffar: Do we really need to create a map with all the headers in order to look for one? Do we want to use the headers for other work?

kalaspuffar:
Hi @alexhauser 

Great work, accept for the little comment I'm ready. Tested and reviewed the code. Also found a small hint that I totally forgot about how the system should work that made some strange errors in the plugin code make sense :)

Also got the feeling this isn't right. Then checked the specification and thought, OK he was right and then finally I realized that I had written the code and forgot about it :)

Best regards
Daniel

app/src/main/java/org/ea/sqrl/processors/CPSServer.java
alexhauser: Only searching for "origin:" within the whole string would be too much of a messy hack in my opinion. So we'd have to iterate over the headers anyway, since we don't know the exact position of the wanted header. So why not do it properly in the first place?

I prefer writing readable, self-documenting and maintainable code over over-optimizing. I once read in a paper one should write nice code and only start optimizing if there are actual performance-problems. This struck a chord with me.

I would even go so far as to say that all the ad-hoc HTTP request parsing code which is present in multiple places within the project at the moment should be broken out into a common request parsing class. This would vastly increase maintainability and readability of the codebase.

But that's only my personal preference, and I have no problem at all with following the project lead when it comes to questions like that. So if you have a concrete suggestion on how to improve this part of the code, please let me know and I'll implement it that way! 😁

kalaspuffar: Thanks for the lengthy comment and I think there is something to the point. I've optimized code to early many times and I've worked with memory constrained environments.

And much of the optimizations can be handled by the compiler. So for this code it does not matter but for the future we need to keep in mind that I want the project to work on ANY phone that uses API 16 and above.

We want a broad reach so can we constrain the memory usage we should. But I'm not one to argue over a few bytes :)

alexhauser: Great discussion, and point taken on the topic of keeping memory usage for older device down! :)

PR #394, opened 2019-07-23
Improve layout of CPSMissingActivity

alexhauser:
**Description:**

Improve the layout of `CPSMissingActivity`.

This is how the activity looks before the proposed changes:

... and here with the changes applied:


sesam:
Good improvement! UX could be further polished, but probably best to keep the momentum and do such things in a later round, maybe inviting some experts -- to avoid changing translatable texts several times.

Issue #393, opened 2019-07-23
Discussion: Rename SimplifiedActivity


alexhauser:
### Enhancement:

Now that we've removed a lot of the legacy activities and code, I think it would be appropriate to rename `SimplifiedActivity` to something that better fits its current role.

I would suggest renaming it to `MainActivity`.

kalaspuffar:
Hi @alexhauser 

Sounds like a great idea. We don't need to keep legacy, and this is the main activity that you will go to on launch most of the time.

But open to suggestion on other names. Any suggestions @sengsational or @sesam ?

Best regards
Daniel

sesam:
> There are only two hard things in Computer Science:
> 1. Naming things...
> 

Jokes aside, the suggested MainActivity looks good to me :)

PR #398, opened 2019-07-24
Rename SimplifiedActivity and remove unused tests.

alexhauser:
Issue #393.

**Description:**

Rename `SimplifiedActivity` to `MainActivity` and remove accessibility tests for previously removed activities.


app/src/main/java/org/ea/sqrl/activites/MainActivity.java
```diff
@@ -79,7 +79,7 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         IntentResult result = IntentIntegrator.parseActivityResult(requestCode, resultCode, data);
         if(result != null) {
             if(result.getContents() == null) {
-                Log.d("SimplifiedActivity", "Cancelled scan");
+                Log.d("MainActivity", "Cancelled scan");
```
kalaspuffar: This log row should use TAG constant as all other log rows :)

alexhauser: Yep, I'll fix this right away.

alexhauser:
Hi @kalaspuffar , 

the log tag issue is resolved now, should be ready to merge.

Cheers, Alex

Issue #396, opened 2019-07-23
Improve program flow on "import with recovery"


alexhauser:
### Issue or error solving.

From https://sqrl.grc.com/threads/first-run-import-worked-but-didnt-clear.833/

> I've just downloaded and run the app for the first time and opted to import the identity using a qr code and the recovery key. This seemed o work fine but after creating a password and setting a friendly name was sent back to the "how do you want to add your identity" page. I assumed that something had not worked so did the process a second time and was again sent back to the "add new id" page. This time round I quit the app and re-launched it and was presented with the "scan qr code for site" page, which is what I was expecting after importing my id.

> Seems to me that the add and id page should have a cancel / I'm done button or that during the "first run" it goes to the import id page only once.

gwes:
It was me who dropped this on the forums, feel free to contact me on here if you need any more info.

alexhauser:
Hi @kalaspuffar ,

how should we go about this. The last activity in the "import with rescue code" flow is `RenameActivity`, which we currently just `finish()` after doing the renaming. A quick fix would be to do a `finishAffinity()` and then launch the Simplified/Main activity.

However, we call `RenameActivity` from different places and I'm not sure if this procedure is appropriate for all of them.

What's your take on this?

Best regards,
Alex

alexhauser:
I've tested this and found that the approach described above should work just fine. Going to post a PR now.

PR #409, opened 2019-08-01
Improve program flow on "import with recovery"

alexhauser:
Issue #396

**Description:**

Improve program flow on "import with recovery" by launching the main activity after the "rename" step.


kalaspuffar:
Hi @alexhauser 

Seems to work perfectly. Could be a bit irritating if you need to rename multiple identities but that is an edge case so we should not focus on that.

Best regards
Daniel

Issue #404, opened 2019-07-25
Improved clarity on password usage.


kalaspuffar:
### Feature request.

*Short description of the feature*

suggestion from user: 
provide more context for what password the app requests when the user attempts to log into a website. is this a website specific password or an SQRL identity password or SQRL vault (app) password or what?

alexhauser:
Valid point. Should be easily solvable by adding another info icon to the right of the password field which will show a help text regarding the identity password when it's being tapped on.

Or does anyone have a better implementation idea?

sengsational:
This was my very first observation when trying the app for the first time (https://sqrl.grc.com/threads/just-a-ramble-on-my-first-day-with-sqrl.336/).

>  The app flipped back out of camera mode. Then it's asking for a "password". Here I found myself wondering "does this app think I already have a password for sqrl.grc.com? And I'm thinking "whoa, I'm not giving my SQRL master password to this hinky forum site, who knows who this Gibson character is, anyway" ;) Seriously, the point I'm trying to make is that instead of "password", it should probably be labeled "SQRL Master Password" and have an info thing that pops up and says "your master password is used to decrypt your identity...it won't leave your phone". 

I like the idea of a more info user action that will clarify what's happening with the password.

alexhauser:
OK, so we're talking about two different scenarios here: 

The OP was referring to the password field on the login screen, while @sengsational is talking about creating the identity password in `SaveIdentityActivity`.

Both of those should be addressed, I guess.

PR #410, opened 2019-08-01
Add help for password creation/usage

alexhauser:
Issue #404

**Description:**

Add info icons providing help texts regarding password creation/usage on `SaveIdentityActivity` and `LoginActivity`.


kalaspuffar:

I chose to change the help text for the password to "Please provide the full-length SQRL identity password that you have chosen for this identity."

Maybe you have a better suggestion. But I don't want to create confusion for our end users. The old phrase said to use the password chosen at creation but you can change and reset your password so we might want to keep it simpler?

Thoughts, suggestions?

Best regards
Daniel

alexhauser:
Hi @kalaspuffar,

I agree and like your revised version!

Cheers, Alex

Issue #414, opened 2019-08-05
Simplify naming when creating a new identity


alexhauser:
### Enhancement:

In issue #405, @kalaspuffar answered regarding the naming of identities:

> First identity get the name default.

This is true only when the first identity is being **imported**. The situation is different when the first identity is being created from scratch. In this case, we currently always ask for an (optional) identity name:

<img src="https://user-images.githubusercontent.com/4005543/62491778-2c3ec280-b7cd-11e9-8ee3-df43e8761e05.jpg" width="300">

My suggestion would be to completely remove the *Identity name* field from `SaveIdentityActivity` and handle naming like we do it on import: If it's the first identity, just silently use the (new) default identity name, and if not, then send them over to the `RenameActivity` in the last step.

What do you guys think?

kalaspuffar:
Hi @alexhauser, @sesam and @sengsational 

I think this change is reasonable. If we feel different in the future we could move to another more streamlined process but removing the optional name from the creation process might be something that will increase adoption as every extra introduced step will make people hesitant. For instance:

"I want to pick a name but don't know what is appropriate but it's optional so I might want to or should I?"

If the option is not presented we give the user a more streamlined experience and they can later find that you can actually rename it if you like.

Then again for more experienced users we might have the "problem" that they have a name already in another client and import it. Then they want to keep their name from the other client. But we already have that issue so it won't change with this change.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar ,

thanks for your feedback, you summarized exactly what I was thinking as well. 

As for the "power user" case, it would be easy to add a global settings toggle like "*Always offer naming identities*", which we would check upon import/creation and if it's activated we always launch the `RenameActivity`.

Best regards,
Alex

sengsational:
Default name (streamlined process) on both new creation and import sounds like the best approach. 

I'm not sure a setting is necessary; if they're a power user, they'll know how to rename.  As soon as they have two or more identities, the application should make it obvious that renaming is an option.  I haven't played with multiple identities in a long time, so just talking conceptually.

PR #416, opened 2019-08-06
Simplify naming when creating a new identity

alexhauser:
Issue #414.

**Description:**

When creating a new identity, don't ask for an identity name if it's the first identity. 
(In essence, duplicate the behaviour that we use on import)

**Changes:**
- Remove identity name field from `SaveIdentityActivity`
- Upon identity creation, if it is not the first identity, send the user to the `RenameActivity` after saving the identity. Using an intent extra, let `RenameActivity` know that it should launch `NewIdentityDoneActivity` after the renaming process instead of `MainActivity`.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/create/SaveIdentityActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/create/SaveIdentityActivity.java
@@ -11,6 +11,7 @@ import android.widget.ImageView;
 
 import org.ea.sqrl.R;
 import org.ea.sqrl.activites.base.LoginBaseActivity;
+import org.ea.sqrl.activites.identity.RenameActivity;
 import org.ea.sqrl.processors.SQRLStorage;
 import org.ea.sqrl.utils.PasswordStrengthMeter;
 import org.ea.sqrl.utils.SqrlApplication;
@@ -21,6 +22,8 @@ import org.ea.sqrl.utils.SqrlApplication;
  */
 public class SaveIdentityActivity extends LoginBaseActivity {
     private static final String TAG = "SaveIdentityActivity";
+
+    private boolean firstIdentity = false;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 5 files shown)


alexhauser:
I've only posted this as a draft pull request since @sengsational and @sesam have not given their feedback on issue #414 yet.

app/src/main/java/org/ea/sqrl/activites/identity/ImportActivity.java
```diff
@@ -115,8 +115,6 @@ protected void onCreate(Bundle savedInstanceState) {
 
                     if(newIdentityId != 0) {
                         if(firstIdentity) {
-                            mDbHelper.updateIdentityName(ImportActivity.this, newIdentityId,
```
kalaspuffar: Why did we remove this? Shouldn't we set the first identity name to "My identity"?

app/src/main/java/org/ea/sqrl/activites/create/SaveIdentityActivity.java
```diff
@@ -78,19 +80,29 @@ protected void onCreate(Bundle savedInstanceState) {
...
                     txtNewPassword.setText("");
                     txtRetypePassword.setText("");
 
                     SaveIdentityActivity.this.finish();
-                    startActivity(new Intent(this, NewIdentityDoneActivity.class));
+
+                    Intent nextActivity = null;
+
+                    if(firstIdentity) {
+                        nextActivity = new Intent(this, NewIdentityDoneActivity.class);
```
kalaspuffar: Should we add code to set the standard name here?

app/src/main/java/org/ea/sqrl/activites/identity/ImportActivity.java
alexhauser: Since the previous PR, this happens already in `newIdentity()`, so it is already taken care of at that point.

app/src/main/java/org/ea/sqrl/activites/create/SaveIdentityActivity.java
alexhauser: Same as above:

> Since the previous PR, this happens already in `newIdentity()`, so it is already taken care of at that point.

sengsational:
I like this implementation.  Great work Alex!

alexhauser:
Thanks, Dale! 
It seems everyone is on board with this one, so I guess it can be reviewed/merged now.

Issue #419, opened 2019-08-09
New back button


kalaspuffar:
### Feature request.

After seeing the back button we implemented for the Identity Management page. I think we should fix so we have a consistent layout. So where ever applicable we should implement the new way of enabling back for users.

I believe that the back button is optional on devices so giving the user a consistent alternative to going back to the previous page seems like a good addition. What do you think @sesam, @sengsational and @alexhauser?

Best regards
Daniel

sengsational:
I'm thinking that rather than do each activity, we can add some code to `CommonBaseActivity` that would put the 'back' arrow in the title bar.

I added this to `CommonBaseActivity.onCreate()`:

        ActionBar actionBar = getSupportActionBar();
        if (actionBar != null) {
            actionBar.setDisplayHomeAsUpEnabled(true);
            actionBar.setDisplayShowHomeEnabled(true);
        } 

And added this method to `CommonBaseActivity`:

    @Override
    public boolean onSupportNavigateUp(){
        onBackPressed();
        return true;
    }

And it seemed to do what we want: back arrows that don't depend on the phone's back button.  We'd probably want to turn it off on the main screen, though.

kalaspuffar:
Hi @sengsational 

Great suggestion. I want to remove them from screens that are a part of guides as well. Or they need to jump the user back to the right screen, outside of the guide. 

For instance, if you have started an identity creation you can't back to the previous step you need to start over.

Other than that you can create a pull request with this suggestion if you like :)

Best regards
Daniel

alexhauser:
Yes @sengsational, if we really do want a back button on every activity, that would be the way to do it. 

I don't really have a strong opinion on the topic, so I'm gonna let you guys decide how to go on with this.

Oh and by the way, Dale, your screenshot tells me that you're a few commits behind the current status quo, so you might consider pulling all the latest changes before posting a PR.

Cheers, Alex

sengsational:
I'll take a look and see if I can find all the instances of "back" functionality in the guides.  And certainly will grab the most up-to-date code before making any significant changes...my original post here was just a "quickie" to see if it would work.

PR #420, opened 2019-08-12
back arrow on common base

sengsational:
**Issue #419**

**Description:**
Configure `CommonBaseActivity` to have the back navigation in the action bar

**Changes:**
Changed settings for action bar.  Added implementation to call `onBackPressed()` when navigation activated.  Removed `parentActivityName` from identity management manifest item.

**Discussion:**
There's more ways to get this same result than the way this PR does it, but if this does what we want, it's pretty simple.  It seems that if various activities customized `onBackPressed()`, then this solution should honor those customizations (although I'm posting this without very much testing to validate those).  There were no other instances of `parentActivityName` associated with activities in the manifest, but that would also be an alternative way to approach this.

In cases like this, it would be helpful to have a diagram that showed all the linkages between activities and how they should operate, for example, in some cases, the previous activity makes no sense, so we must start from an earlier point.  But I'm not quite familiar enough with all of the use-cases to know which activities are of this specialized type. 

I did see a `Close` button, which could have been removed, but I didn't know if removal of that (and similar) was something we wanted to do, or if keeping the button would make it more clear for the users.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -31,7 +31,6 @@
         <activity
             android:name=".activites.identity.IdentityManagementActivity"
             android:label="@string/title_identity_management"
-            android:parentActivityName=".activites.MainActivity"
             />
 
         <activity
--- a/app/src/main/java/org/ea/sqrl/activites/MainActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/MainActivity.java
@@ -20,6 +20,7 @@ import org.ea.sqrl.utils.SqrlApplication;
 import org.ea.sqrl.utils.Utils;
 
 import java.util.Arrays;
+import java.util.Objects;
 
 /**
  *
@@ -36,6 +37,9 @@ public class MainActivity extends LoginBaseActivity {
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_main);
+
+        Objects.requireNonNull(getSupportActionBar()).setDisplayHomeAsUpEnabled(false);
+        Objects.requireNonNull(getSupportActionBar()).setDisplayShowHomeEnabled(false);
 
         rootView = findViewById(R.id.mainActivityView);
         communicationFlowHandler = CommunicationFlowHandler.getInstance(this, handler);
--- a/app/src/main/java/org/ea/sqrl/activites/StartActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/StartActivity.java
@@ -14,6 +14,8 @@ import org.ea.sqrl.activites.identity.ImportActivity;
 import org.ea.sqrl.activites.identity.ImportOptionsActivity;
 import org.ea.sqrl.activites.identity.TextImportActivity;
 
+import java.util.Objects;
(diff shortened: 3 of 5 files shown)


alexhauser:
Does this work to go back and retain all the relevant data in our guides/ wizards? (like create identity, rekey etc.?)

So for example, can I go back from confirming/entering the rescue code to showing it again in the "create identity" wizard?

sengsational:
I was not sure what "guides" meant in #149, but with the addition of the description "wizard", above, I now see that it's the process of creating or importing an identity.  I did not test the before/after this change in those use-cases, but that needs to be done before going forward with this.

sengsational:
> So for example, can I go back from confirming/entering the rescue code to showing it again in the "create identity" wizard successfully without losing data?

@alexhauser, I tested this particular thing (hitting 'back' from `RescueCodeEnterActivity`).  With either the phone's back button or the newly added button on the action bar, it works the same both ways.  It goes all the way back to `IdentityManagementActivity` or `StartActivity` depending on which one it started from.  This agrees with Daniel's comment from #419:

>For instance, if you have started an identity creation you can't back to the previous step you need to start over.

@kalaspuffar , when you say "I want to remove them from screens that are a part of guides as well. Or they need to jump the user back to the right screen, outside of the guide", I'm not sure what activities you're referencing.

I went through what I thought were all use-cases (except for `TextImportActivity`, which would be too tedious).  In many places, I tried the phone's back button and compared the functionality to the left arrow on the action bar.  In all cases that I tested, the behavior was identical.  I noticed a "close" button, which had the same effect, but wasn't sure that needed to be removed or not.

alexhauser:
Hi @sengsational, @kalaspuffar and @sesam ,

to me, showing a "back arrow" heavily reinforces the assumption for the user that it's possible to go "one step back" within the given wizard. 

And let's be honest: We should make this possible. What good is it to probe if the user has correctly written down the rescue code, if we're not letting him correct his error should there be one.

I think that "enable going back" was also something that was mentioned in the "UX help" thread in the forums before everyone abandoned ship.

I guess this warrants its own issue, though. What do you guys think?

Best regards,
Alex

sengsational:
Yes, Alex, I think you're right about user expectations of 'one step back'.  So although adding the left arrow (this PR) ends with the same result as the phone's back button, I can see how pressing the left arrow might cause slightly more of a 'surprise' than pressing the phone's back button.  This is because users might expect more 'drastic' actions from using that form of 'back'.

I do think that stepping back in smaller increments probably should have it's own separate issue.  When I experienced jumping far back in identity creation, I thought it might have security implications, and those, if any, could be discussed.

This PR could be closed and made part of the proposed additional issue, or go forward as-is.  I'd vote to just let it go as-is.

kalaspuffar:
Hi @alexhauser 

I think the idea is worth revisiting. I modeled the wizards to only allow one-way flow in order to keep a secure transaction. Many key generation flows are one-way and we are also dealing with something that the user needs to record correctly and should not be able to workaround.

We want to avoid someone remembering 4 number going forward, typing in, going back remember 4 more and so on.

Either way, this should be discussed in a separate issue and I will concentrate on testing and accepting this PR as is.

Best regards
Daniel

Issue #421, opened 2019-08-13
Ability to navigate to identity management parent from identity management child activities


sengsational:
### Identity Management shows in the options menu in child activities

*Short description of the issue*
The options menu allows users to go to "identity management" from activities which are spawned from the activity management activity itself. 

*Expected behavior*
Users would not be given the opportunity to go to identity management if they were already performing some identity management activity.

*Actual behavior*
Users can select "identity management" from export identity activity, for example, then the activity stack gets confused.

*Steps to reproduce*
For example, go into export identity, then use the options menu to select "identity management".  Now it the 'back' button on the phone or in the action bar.  Focus will return to the export identity (weird).  If the user was not given the opportunity to go to the base identity management from a child activity management activity, then this wouldn't happen.

*Proposed solution*
Remove "identity management" from the options menu in child activities.  In `BaseActivity.onCreateOptionsMenu()`, we already have `action_identity_management` set to not visible.  Activities that are considered children to identity management should also be included there.

alexhauser:
I fully agree!

PR #422, opened 2019-08-14
Selective display of identity management in options menu

sengsational:
***Issue #421***

***Description:***
Display options menu on `MainActivity` only.

***Changes:***
Altered visibility of `action_identity_management` options menu item to be gone from everywhere except the main  `MainActivity`

***Discussion:***
Almost all of the activities that extend `BaseActivity` (where the options menu is defined) are "downstream" in activity flow of `IdentityManagementActivity`, so don't have the need to navigate to identity management this way.  There might be a few non-downstream activities, such as `LoginBaseActivity` and `StartActivity` that, before this change had the ability to navigate to identity management, but it didn't seem to make sense to add those (but we could).

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/base/BaseActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/base/BaseActivity.java
@@ -138,7 +138,9 @@ public class BaseActivity extends CommonBaseActivity {
             menu.findItem(R.id.action_language).setVisible(false);
         }
 
-        if (this instanceof IdentityManagementActivity) {
+        if (this instanceof MainActivity) {
+            menu.findItem(R.id.action_identity_management).setVisible(true);
+        } else {
             menu.findItem(R.id.action_identity_management).setVisible(false);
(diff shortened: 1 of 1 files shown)


sengsational:
sqrlAndroidActivityFlow

alexhauser:
I think it's reasonable to only show the identity management menu item in `MainActivity`. So for me, this one is good to go.

sesam:
Nice flowchart! Is this added to the app documentation somewhere?

kalaspuffar:
> Nice flowchart! Is this added to the app documentation somewhere?

You are right @sesam this is should be added to the docs. In the doc folder today we only have my 
http://draw.io diagram with all the activities and their interaction and I believe it's very inaccurate after all our changes.

I've added @sengsational new flow chart, and maybe someone will have the time later to update the draw.io version so we can have something that can be updated when changing data.

Perhaps there is a better tool; I used draw.io because it was easy :)

Best regards
Daniel

Issue #438, opened 2019-09-26
Update rescue code page to new design.


kalaspuffar:
The rescue code page does not have the explaination why we should keep the page.

alexhauser:
Hi @kalaspuffar,

are we talking about the print version of the rescue code page here?

kalaspuffar:
Hi @alexhauser 

Correct, I saw Steve's new layout during the Meetup here in Sweden and I did not want to interrupt the talk with a lengthy report but still remember it as a to-do.

The print page today is only the code in red which is a bit sparse :)

Best regards
Daniel

alexhauser:
Ok, thanks! I'll look into this.
Cheers, Alex

alexhauser:
I basically copied Steve's layout and texts, but also added our existing footer design:

![rescue_code_print](https://user-images.githubusercontent.com/4005543/65764925-1b546280-e127-11e9-8493-1faec6796bcc.jpg)

Any suggestions or change requests?

kalaspuffar:
Hi @alexhauser 

Looks great! I just want us to be compliant with the reference client and make it clear to users that this page is important. Our earlier design was quite lacking. 

Great work.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar,

I totally agree! 

I'll clean things up and post a PR asap. In the PR we can discuss the actual implementation (I tried to avoid duplicating code from `IdentityPrintDocumentAdapter` and instead breaking everything out into reusable chunks) - but let's talk about that when I post the actual code.

Best regards,
Alex

PR #439, opened 2019-09-27
Improve rescue code printout page

alexhauser:
Issue #438.

**Description:**
Improve the layout and UX of the page that is being created when a user prints the rescue code during the creation of a new identity.

This basically duplicates the layout and texts from GRC's reference client and adds an additional app-specific footer.

**Changes:**
- Add headline, warning and descriptive texts
- Add footer (logo and app info)
- Break drawing the footer layout out into a utility function `Utils.drawPrintPageFooter(...)` so that it can be reused by by all `PrintDocumentAdapter`s
- Add utility function `Utils.drawTextBlock(...)` to draw (multiline) text onto the canvas

rescue_code_print

**Discussion:**

@kalaspuffar, I noticed that in `IdentityPrintDocumentAdapter` you created a custom text-wrapping functionality for drawing multi-line text blocks. My new utility function `Utils.drawTextBlock(...)`, which I use in `RescueCodePrintDocumentAdapter`, uses Android's stock `StaticLayout` functionality instead, which - at least in theory - should be more robust than any home-brewn solution. 

If we want to use my new approach in `IdentityPrintDocumentAdapter` as well, it could replace all the custom functionality of `drawBlockOfText(...)`, `findNextString(...)` as well as `drawCenteredText(...)`.

If you agree, I could easily port `IdentityPrintDocumentAdapter` over to the new `Utils.drawTextBlock(...)` in a follow-up PR.

Also, I'm interested in your opinion on whether it's ok to have those utility functions in our general `Utils` class or if you'd prefer having a dedicated class á la `DocumentPrintUtils` instead.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/services/IdentityPrintDocumentAdapter.java
+++ b/app/src/main/java/org/ea/sqrl/services/IdentityPrintDocumentAdapter.java
@@ -1,9 +1,7 @@
 package org.ea.sqrl.services;
 
 import android.app.Activity;
-import android.content.pm.PackageInfo;
 import android.graphics.Bitmap;
-import android.graphics.BitmapFactory;
 import android.graphics.Canvas;
 import android.graphics.Color;
 import android.graphics.Paint;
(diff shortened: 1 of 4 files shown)


kalaspuffar:
Hi @alexhauser 

I think we can separate this concern into ```DocumentPrintUtils``` except for that I think the approach to use standard things are good. I thought it was a simple thing to do and in java, you usually have to do these simple calculations so I did not even look for another solution.

Best regards
Daniel

PR #441, opened 2019-10-10
Fix potential "index out of bounds" in EncryptionUtils.xor()

alexhauser:
**Description:**
In `EncryptionUtils.xor()`, we iterate over two byte arrays, using the length of the first array as the abort condition. Since the second byte array may be shorter, this creates a potential for an `ArrayIndexOutOfBoundsException`.

**Changes:**
Apply a modulo operation on the second array's indexer so that it will wrap around to 0 if the second array is shorter than the first one.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/utils/EncryptionUtils.java
+++ b/app/src/main/java/org/ea/sqrl/utils/EncryptionUtils.java
@@ -219,10 +219,26 @@ public class EncryptionUtils {
         return new String(decodeUrlSafe(data), "UTF-8");
     }
 
+    /***
+     * Performs a byte-by-byte XOR operation on the provided byte arrays.
+     * Since this is being used for XOR'ing cryptographic keys, in order to avoid information leakage
+     * caused by (partial) reuse of key material, this function enforces both byte arrays to have the
+     * same length (nr of bytes).
+     *
+     * @param a The first operand for the XOR operation. Must be same length as b.
+     * @param b The second operand for the XOR operation. Must be same length as a.
+     * @return The byte array resulting from the XOR operation.
+     * @throws IllegalArgumentException If the lengths of a and b are not equal.
+     */
     public static byte[] xor(byte[] a, byte[] b) {
+        if (a.length != b.length) {
+            throw new IllegalArgumentException(
+                    "The byte arrays to be XOR'ed need to have the same length!");
+        }
+
         byte[] result = new byte[a.length];
         for (int i = 0; i < a.length; i++) {
-            result[i] = (byte) (((int) a[i]) ^ ((int) b[i%b.length]));
+            result[i] = (byte) (((int) a[i]) ^ ((int) b[i]));
         }
         return result;
     }
--- a/app/src/main/java/org/ea/sqrl/utils/EncryptionUtils.java
+++ b/app/src/main/java/org/ea/sqrl/utils/EncryptionUtils.java
@@ -223,7 +223,7 @@ public class EncryptionUtils {
      * Performs a byte-by-byte XOR operation on the provided byte arrays.
      * Since this is being used for XOR'ing cryptographic keys, in order to avoid information leakage
      * caused by (partial) reuse of key material, this function enforces both byte arrays to have the
-     * same length (nr of bytes).
+     * same length (number of bytes).
      *
      * @param a The first operand for the XOR operation. Must be same length as b.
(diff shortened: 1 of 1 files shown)


sesam:
Travis failed here - I had a look at the logs and found

> <============-> 96% EXECUTING [10m 8s]> 
> Task :app:connectedDebugAndroidTest FAILED
> <============-> 98% EXECUTING [10m 8s]
> Execution failed for task ':app:connectedDebugAndroidTest'.

which maybe is a known issue with Travis.

alexhauser:
Hi @sesam,

the travis build has been failing for quite some time now, @kalaspuffar knows about it but I think this has low priority right now. What I'm trying to say is that this has nothing to do with the current PR but is a general problem of the build script.

Best regards,
Alex

kalaspuffar:
Hi @alexhauser 

Great suggestion and you are right that this fix would solve the problem in question. And it would be really bad if the second length was longer because then some of the key material would not be xor:ed at all.

In the current implementation, though, we only use the xor function to xor keys with the same length over and over. So the overflow check is not needed in this particular case.

The other reason I don't want to merge this pull request is that if we happened at the case where the key was shorter than the buffer length or vise versa would mean that you either xor bytes multiple times or reusing the key which is strictly prohibited.

In crypto, that could make the crypto less useful.

Let's, for instance, say that we want to encrypt a text of 100 bytes using xor, and we have a 5 bytes key. That would mean that we reused the same 5 bytes 20 times. This could lead to an issue where you could easily find out the 5 bytes.

On the other hand, if you do the same operation and create a key of 100 bytes, then it would be improbable to figure out the original text from the ciphertext without the key as every byte has been altered by a different key byte.

I hope I make my point clear. If you have any questions, please don't hesitate to ask.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar,

thanks for your feedback! I fully understand your reasoning and agree that my proposed solution - while solving the problem at hand - could lead to undesired effects if someone doesn't know what he's doing. So I fully agree that the PR should not be merged as-is!

That being said, I'm still unconfortable with the fact that a public utility function with the rather generic name of `xor(...)` , which may be used by an unwitting contributor in the future for a different purpose than XORing keys, will lead to an unexpected crash caused by an `ArrayIndexOutOfBoundsException`. I feel like having a function crash unexplicitly is bad practice, and looks like a coding oversight to anyone reading/auditing the codebase.

What do you think about enforcing equal lengths of the two byte arrays instead, and giving a good error message in case they are not? Something like...

```java
if ( a.length != b.length ) {
    throw new IllegalArgumentException(
        "The byte arrays to be XORed need to have the same length!");
}
```
or, alternatively, just assert length equality:
```java
assert a.length == b.length; 
```

Both solutions would make everything more explicit while avoiding the crypto pitfalls which you've described earlier. What do you think?

kalaspuffar:
Hi @alexhauser 

I like solution number 1 because then we explicitly tell implementers what's going on and why we restrict the function to be used with only arrays of the same length.

Perhaps we could add a useful JavaDoc comment aswell describing the functionality and the limitation we put on this implementation and why.

Best regards
Daniel

alexhauser:
Perfect @kalaspuffar, I'll post a followup commit asap.

alexhauser:
Hi @kalaspuffar, 

I've just pushed a commit with the discussed changes, looking forward to your review.

Cheers, Alex

app/src/main/java/org/ea/sqrl/utils/EncryptionUtils.java
```diff
@@ -219,7 +219,23 @@ public static String decodeUrlSafeString(String data) throws Exception {
         return new String(decodeUrlSafe(data), "UTF-8");
     }
 
+    /***
+     * Performs a byte-by-byte XOR operation on the provided byte arrays.
+     * Since this is being used for XOR'ing cryptographic keys, in order to avoid information leakage
+     * caused by (partial) reuse of key material, this function enforces both byte arrays to have the
+     * same length (nr of bytes).
```
kalaspuffar: Don't think we need to abbreviate (number of bytes)

alexhauser: Fixed. Should be ready to merge now.

alexhauser: Fixed. Should be ready to merge now!

PR #445, opened 2019-11-16
Fixing the rescue code creation.

kalaspuffar:
Description:
Fixing the rescue code generation.

Changes:
Added a description and changed to use a 32-byte random source.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/processors/SQRLStorage.java
+++ b/app/src/main/java/org/ea/sqrl/processors/SQRLStorage.java
@@ -102,7 +102,7 @@ public class SQRLStorage {
     }
 
     public void newRescueCode(EntropyHarvester entropyHarvester) {
-        tempRescueCode = new byte[12];
+        tempRescueCode = new byte[32];
         entropyHarvester.fetchRandom(tempRescueCode);
     }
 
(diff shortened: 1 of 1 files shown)


alexhauser:
Hi @kalaspuffar ,

I have looked into this a little closer and wrote some tests regarding rescue code digit distribution. I compared our current implementation with Paul Farrer's Golang version, and both perform very well:

rc_digit_distribution

Just for the record, here is my testing code:
I ran this twice, the first test used our current code, and for the second test I ported over Paul Farrer's Go code and used it in `getTempRescueCode()`

``` java
SQRLStorage storage = SQRLStorage.getInstance(this);
Map<Character, Long> dist = new HashMap<>();

for (int i=0; i<500000; i++) {
    try { storage.newRescueCode(EntropyHarvester.getInstance()); }
    catch (Exception e) { e.printStackTrace(); }
    String rc = storage.getTempRescueCode();

    for (int y=0; y<rc.length(); y++) {
        Character c = rc.charAt(y);
        if (!dist.containsKey(c)) dist.put(c, 0L);
        dist.put(c, dist.get(c) + 1);
    }
}
```
With these reassuring results, I think we can move on and stop worrying about our small divergence from the spec.

Best regards,
Alex

kalaspuffar:
Hi @alexhauser 

Wonderful, it's one thing to believe, another to test. I'm glad you took the time to verify the distribution.

Best regards
Daniel

Issue #452, opened 2020-01-03
EnHash regression


alexhauser:
### Issue or error solving.

When implementing the unit test vor the enhash-vectors (see #448), I noticed that **all of the tests were failing**.

It seems that this is caused by a regression introduced by commit [e49cd4c](https://github.com/kalaspuffar/secure-quick-reliable-login/commit/e49cd4cc2a18ea3a5902094111c55a7792d57004) created on Nov 4th, 2019. 

This is somewhat serious, since all the identities created using that code would technically be broken, since the IUK <-> IMK/ILK relationship is invalid, and a text export and import into another client would produce a different IMK/ILK. 

Fortunately, as far as I can tell, we have not released anything containing that flawed code, right?

PR #453, opened 2020-01-03
Fix EnHash regression

alexhauser:
Issue #452.

**Description:**
Fix a regression where the number of hash iterations within `EncryptionUtils.enHash()` was increased by one, causing the function to produce incorrect results.

Increasing the iteration count from 15 to 16 within the for-loop was unnecessary, since the first iteration is already done outside the loop.


kalaspuffar:
Hi @alexhauser 

Thank you for catching this one. You usually think correctly the first time when you have all the context.

I should have checked the code another time before this change.

Best regards
Daniel

Issue #457, opened 2020-01-15
Error when leaving "QuickPass Timeout" field empty


alexhauser:
# Issue or error solving.

See https://sqrl.grc.com/threads/ui-bug-not-clean-after-error.1060/

> SQRL 1.5.1, Android 7.0
>1. Identity Management
>2. Settings
>3. Delete the time from the QuickPass Timeout.
>4. "Save", then enter password.
>5. "Encrypting Identity" message box is displayed, then
>6. Error message -- "QuickPass timeout input is not a number"
>7. "OK" to acknowledge
>8. The "Encrypting Identity" box and darkened screen persist, cannot dismiss to allow interaction with the Settings page
>9. Must use back button to return to previous screen (Identity Management home page), then go to Settings again
I would expect that once the error is acknowledged, the messages should clear and allow the user to then directly interact with the Settings page.

PR #460, opened 2020-01-25
Add input validation to identity settings

alexhauser:
Issue #457.

**Description:**

Add input validation to all `EditText` input fields in the identity settings activtiy. This fixes a bunch of issues which could be caused by either leaving input fields empty or entering very large values.

**Changes:**

- Before, the few validations present were done upon hitting the "Save" button. Now, input validation is done "on-the-fly", directly when entering the values.
- The "Save" button is now dynamically enabled/disabled according to the input validation state
- Input errors are displayed directly on the affected input field instead of in a popup message
- Remove the arbitrary _QuickPass Timeout_ limit of one week

The whole process should be much cleaner now.

**Validations that are performed now:**

- Check for empty field
- Check for non-numeric values
- Upper bounds checking


app/src/main/java/org/ea/sqrl/activites/identity/IdentitySettingsActivity.java
```diff
@@ -110,25 +97,83 @@ public void onResume() {
...
-        cbSettingsNoBypass.setChecked(storage.isNoByPass());
+        txtQuickPassLength.setText(Integer.toString(storage.getHintLength()));
+        txtPwdVerifySecs.setText(Integer.toString(storage.getPasswordVerify()));
+        txtQuickPassTimeout.setText(Integer.toString(storage.getIdleTimeout()));
+        cbSQRLOnly.setChecked(storage.isSQRLOnly());
+        cbNoBypass.setChecked(storage.isNoByPass());
     }
 
-    public int getIntValue(EditText txt, int errorMessage) {
+    public int getIntValue(EditText txt) {
```
kalaspuffar: Is -1 a valid number? It is below 65535 at least.

alexhauser: This method always returned -1 on error, I haven't changed anything there. I've just removed the call for showing the error message. Or am I missing your point?

alexhauser: If you're talking about lower bounds checking on the input fields, that would need to happen in `validateInputFields()`. And yes, I've thought about it too. I guess we should only allow positive integers. Right? If you agree, I'll modify my code to accomodate that.

kalaspuffar: Yes, that was what I was thinking about. Might have expressed it a bit bad.

alexhauser:
Added lower bounds check - please review again, @kalaspuffar. 

It technically isn't really necessary, since the input field doesn't let you input negative numbers, but hey - belts and suspenders. 😀

Issue #435, opened 2019-09-15
Introduction Guide / Wizard


kalaspuffar:
### Feature request.

*Short description of the feature*
Creating a short introduction wizard/guide that will replace the block of text we have on our introduction page.

I sat down today and created some screens that explain my thoughts. I'm no designer so the balance of the pages is all over the place but I hope that the general idea is clear enough.

![page1](https://user-images.githubusercontent.com/1424084/64926619-ee877d80-d7ff-11e9-9be8-3f1052242aaf.png)
![page2](https://user-images.githubusercontent.com/1424084/64926620-ee877d80-d7ff-11e9-811b-2e03e64b2ae3.png)
![page3](https://user-images.githubusercontent.com/1424084/64926621-ee877d80-d7ff-11e9-8e0d-0b58b09b4fdb.png)
![page4](https://user-images.githubusercontent.com/1424084/64926622-ee877d80-d7ff-11e9-8894-3b18210f3503.png)

Thoughts @sengsational, @sesam and @alexhauser?

Maybe someone with more graphical skills could add some illustrations to convey the information in a more instructive way. Perhaps we want different introduction bullets or in a different order.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar,

I really like the idea of breaking up the huge block of intro text into only a few nice looking, small and easily digestable snippets.

I also like the general look-and-feel of your drafts. I think if we had someone to give it a final touch-up and maybe even add some vector art, this could be a perfect onboarding experience.

sesam:
This feels like a great idea! I think it's worth to take some iterations on the texts and ordering too. My first impression was WOW! Reading it two more times, I'm starting to think how it could be reorganised to keep what users care about up front. Maybe also make it more "selling" and keep what's meant for more technically minded people later in. So those who want to get started can learn just a little bit then get going, while curious know-it-alls go all the way through and get more details.

We could use the time when the user is reading to prepare other things behind the scenes. Maybe create and store the first identity with strong = slow encryption.
The other, very outdated, android SQRL app creates an identity before you reach the first screen, and crypto-wallets often create keys automatically, and then nag the user to create a backup, or even make it impossible to use the wallet without first making a backup, the proper way.

[...]

PR #469, opened 2020-03-18
First implementation of startup guide.

kalaspuffar:
Issue #435.

Description:
This is my first implementation of some new startup screens with less text and more graphics, welcoming you into the application.

Changes:
Added wizard screens and removed the old startup screen. Kept the startup screen logic to make the transition as easy as possible.
Screenshot_20200318-230728
Screenshot_20200318-230806
Screenshot_20200318-230814
Screenshot_20200318-230824
Screenshot_20200318-230831
Screenshot_20200318-230836
Screenshot_20200318-230840


kalaspuffar:
Hi @alexhauser 

As you are a native German speaker and the Germanic languages, Swedish included could be hard to handle in text as some words may wrap strangely and so on could you create and test a German version before I merge this change?

I've tested the swedish translation and that seems to work. Will try the Mandarin one today as well.

Best regards
Daniel

alexhauser:
Sure, I'll give it a go as soon as I have some time at hand.

alexhauser:
@kalaspuffar, German seems to work just fine on decent resolutions. Here is the German translation if you want to try it yourself, or if you want to include it into the merge:

```XML
<string name="intro_1">Willkommen zu</string>
<string name="intro_2">SQRL ist _eine_ Identität.\n\nEs erstellt _eindeutige_ Schlüssel für jede Webseite.\n\nUnd enthüllt keine _nachverfolgbaren_ Daten.</string>
<string name="intro_3">Diese App ist _Open Source_ Software und steht unter einer _offenen Lizenz_.\n\nMehr dazu im Menüpunkt _Über_ direkt in der App.</string>
<string name="intro_4">Starten Sie, indem Sie eine Identität _erzeugen_ oder _importieren_.</string>
<string name="wizard_skip">Überspringen</string>
<string name="wizard_next">Weiter</string>
```

However, I've noticed that we do have a general problem on small resolutions:

480x800_1

480x800_2

Some kind of scaling would be needed here it seems.

kalaspuffar:
Hi @alexhauser 

Great catch, changing text size due to screen size is a bit tricky, but there is a new auto-scaling feature that seems to work fine except it might create huge text :)

And the first page with so many text boxes is hard to get scaling correctly. Please try again, and if the first page still has a problem, you are welcome to give it a try. I'm not sure how to improve that page.

Best regards
Daniel

PR #480, opened 2020-12-26
Updated to new Android Studio.

kalaspuffar:


app/src/androidTest/java/org/ea/sqrl/AccessibilityInstrumentedTest.java
```diff
@@ -287,18 +287,6 @@ public void testShowIdentityActivityAccessibility() throws Exception {
         onView(withId(R.id.showIdentityActivityView)).perform(click());
     }
 
-    @Test
-    public void testStartActivityAccessability() throws Exception {
```
alexhauser, line 291: Just curious why you removed this test?

alexhauser (approved):
I haven't tested, but looks good.

app/src/androidTest/java/org/ea/sqrl/AccessibilityInstrumentedTest.java
kalaspuffar: Hi @alexhauser 

With the new views for starting using the application this view was removed. Might create a couple of new tests for the new views when ever I get the tests to work again.

Best regards
Daniel

PR #484, opened 2020-12-27
Release 1.7.1

kalaspuffar:
* Enable "show password" option when changing password
* Re-mask passwords when submitting
* Compiled for newer Android versions.


alexhauser (approved):

alexhauser:
@kalaspuffar, we might think about merging #485 and #488 and including those fixes here for the 1.7.1 release.

Issue #481, opened 2020-12-26
QR codes not reading properly


bilde2910:
Some QR codes don't appear to scan properly. See e.g. the code below, which will consistently fail for me:

![image](https://user-images.githubusercontent.com/6617815/103156124-ce39c300-47a5-11eb-9145-03df3a5c6157.png)

The URI encoded is this:  
`sqrl://apps.varden.info/wfptools/history/sqrlauth.php?x=17&sfn=V2F5ZmFyZXIgSGlzdG9yeQ&nut=Uwx16PPuRdxkSwFxQYCBADCjnNAMRl3EKJH6XQkhIaY`

When the app scans the code to authenticate, this is what it somehow ends up with:  
`sqrl://apps.varden.info/wfptools/history/sqrlauth.php?x=17&sfn=V2F5ZmFyZXIgSGlzdG9yeQ&nut=Uwx16PPuRdxkSwFxjnNAMRl3khIaY`

Note the difference in the nut:

```
Original: Uwx16PPuRdxkSwFxQYCBADCjnNAMRl3EKJH6XQkhIaY
Scanned:  Uwx16PPuRdxkSwFxjnNAMRl3khIaY
Spaced:   Uwx16PPuRdxkSwFx       jnNAMRl3       khIaY
```

The "spaced" one is the scanned one, but adding in spaces so that it aligns to the original one. You can tell that it has mysteriously dropped some characters from the base64url-encoded nut.

I have tried switching around the `nut` and `sfn` and if the `sfn` is at the end, then the `sfn` appears to be corrupted in this way. It does not happen to all codes, but it happens to the majority of the codes I generate this way.

Should be noted that when one QR code fails to scan, it appears to fail consistently (dropping the same bits) every time that code is scanned, no matter how close/far away I am or what angle I attempt to scan it from. This would suggest that detects the code correctly, but parses it wrongly.

---

- **Device:** OnePlus 6
- **App version:** 1.7.0 (Google Play)
- **Android version:** 10
- **QR renderer:** phpQRCode (`QRcode::png($url, null, QR_ECLEVEL_M, 6);`).

Not sure what other kind of information you need, so just ask if you want more info, or more samples

alexhauser:
Thanks for the detailled report, we'll look into it as soon as time permits.

alexhauser:
@bilde2910, I'm able to reliably reproduce the bug here at my end, and I think I also know what might be happening. Working on the fix right now.

PR #485, opened 2020-12-28
Mitigation for QR code decoding issue

alexhauser:
This proposes a temporary mitigation for issue #481 and should form the base for a discussion on the topic.

**Description**:

It turns out that our method of reading and concatinating the `SCAN_RESULT_BYTE_SEGMENTS_{x}` within `Utils.readSQRLQRCode(...)` would sometimes leave us with chunks of the SQRL login URL missing, as layed out in issue #481. It seems as those chunks are simply not present in the byte segments contained in the intent that gets passed in to `onActivityResult(...)`. To me, this looks like a bug in either the `zxing-android-embedded` library that we're using, or even in the underlying `zxing:core` library. 

That being said, using `IntentIntegrator.parseActivityResult(...).getContents()`, which returns the decoded QR code as a string, seems to be delivering correct results for non-binary inputs (which all login URLs should be). 

This however does not work for identity QR codes, since they consist of binary data. This seems to be a problem in its own right, given the discussion found [here](https://github.com/journeyapps/zxing-android-embedded/issues/390). So not only do we have the uncertainty of missing byte segements, but also that binary data doesn't seem to be supported anyway. So after all, we might have only gotten lucky with our limited testing of importing identity QR codes up to now.

@kalaspuffar, I'd appreciate a short discussion on the topic before we jump to any conclusions.

**What this PR proposes (as a temporary mitigation):**

This now uses `IntentIntegrator.parseActivityResult(...).getContents()` for login QR codes and only falls back to reading the binary `SCAN_RESULT_BYTE_SEGMENTS_{x}` when dealing with identity QR codes. This *should* at least make the decoding of **login QR codes** much more robust.

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/MainActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/MainActivity.java
@@ -81,33 +81,34 @@ public class MainActivity extends LoginBaseActivity {
     @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         IntentResult result = IntentIntegrator.parseActivityResult(requestCode, resultCode, data);
+
         if(result != null) {
-            if(result.getContents() == null) {
+            final String qrCodeContent = result.getContents();
+
+            if(qrCodeContent == null) {
                 Log.d(TAG, "Cancelled scan");
                 Snackbar.make(rootView, R.string.scan_cancel, Snackbar.LENGTH_LONG).show();
                 if(!mDbHelper.hasIdentities()) {
                     startActivity(new Intent(this, WizardPage1Activity.class));
                 }
             } else {
-                byte[] qrCodeData = null;
-
-                try {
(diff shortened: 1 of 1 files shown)


kalaspuffar:
Hi @alexhauser 

I'm open to a discussion in any form you like around this issue. It seems pretty damming for this library. When I looked into this and created the first implementation of SQRL Login, there were no other libraries available. Perhaps we should do another take on the landscape.

I can see at least 3 different solutions:

* Find another library
* Help them solve the issue in zxing
* Write our own implementation.

Best regards
Daniel

alexhauser:
Right, @kalaspuffar. 

I would suggest to merge this fix for now as a mitigation for #481 and then think about how to approach a general solution for this topic.

So, if you or maybe @JamesonNetworks could test this, that would be great!

alexhauser:
Closing this in favor of #489.

PR #488, opened 2021-02-28
Reload identity onResume

JamesonNetworks:
Issue https://github.com/kalaspuffar/secure-quick-reliable-login/issues/487

Description:
Forces a reload of the identity data on resume of identity management after create identity

Changes:
Loads the current identity ID from context and does a full reload of the identity data through the SqrlApplication class

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/identity/IdentityManagementActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/identity/IdentityManagementActivity.java
@@ -11,6 +11,7 @@ import org.ea.sqrl.activites.base.BaseActivity;
 import org.ea.sqrl.activites.create.CreateIdentityActivity;
 
 import org.ea.sqrl.utils.IdentitySelector;
+import org.ea.sqrl.utils.SqrlApplication;
 import org.ea.sqrl.utils.Utils;
 
 import java.util.Objects;
@@ -57,6 +58,8 @@ public class IdentityManagementActivity extends BaseActivity {
         if(!mDbHelper.hasIdentities()) {
             IdentityManagementActivity.this.finish();
         } else {
+            long id = SqrlApplication.getCurrentId(this.getApplication());
+            SqrlApplication.setCurrentId(this.getApplicationContext(), id);
             mIdentitySelector.update();
         }
     }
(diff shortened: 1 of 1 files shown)


alexhauser (dismissed):
The bug happens because the first thing we do in `CreateIdentityActivity` is to blindly call `cleanIdentity()`, not considering the fact that the create identity wizard can be cancelled at any time during its many steps.

Now, my first instinct for a "proper" fix would have been to move the `cleanIdentity()` call as far back as possible, and only ever use it where we really need to (i.e. when actually touching the identity data), while also handling the reloading of the current identitiy upon cancellation right there, where the context makes sense. This would get us to `RescueCodeShowActivity`, where we first write to the "cleared" identity. Considering the fact that we still have a few other activities in the wizard afterwards before finally saving the new identity, this approach would require each of those activities to check for cancellation, which makes the whole approach moot.

So, I guess your fix is the best that can be done at the moment without rearchitecting the whole monstrosity of the `SQRLStorage` class, which is something that I think we'll have to tackle sometime in the future. 

The only thing that I would like to suggest is adding a short comment to your code, explaining the need to reload the identity data, since it's not exactly obvious at that point.

Thanks again for taking the time and submitting a PR!

JamesonNetworks:
> The bug happens because the first thing we do in `CreateIdentityActivity` is to blindly call `cleanIdentity()`, not considering the fact that the create identity wizard can be cancelled at any time during its many steps.
> 
> Now, my first instinct for a "proper" fix would have been to move the `cleanIdentity()` call as far back as possible, and only ever use it where we really need to (i.e. when actually touching the identity data), while also handling the reloading of the current identitiy upon cancellation right there, where the context makes sense. This would get us to `RescueCodeShowActivity`, where we first write to the "cleared" identity. Considering the fact that we still have a few other activities in the wizard afterwards before finally saving the new identity, this approach would require each of those activities to check for cancellation, which makes the whole approach moot.
> 
> So, I guess your fix is the best that can be done at the moment without rearchitecting the whole monstrosity of the `SQRLStorage` class, which is something that I think we'll have to tackle sometime in the future.
> 
> The only thing that I would like to suggest is adding a short comment to your code, explaining the need to reload the identity data, since it's not exactly obvious at that point.
> 
> Thanks again for taking the time and submitting a PR!

Updated with a comment. Looking forward to helping more! I can't believe I just got around to playing with SQRL. I've been hearing about it for a long time and was excited by the potential.

alexhauser (approved):
Perfect!

alexhauser:
Waiting for @kalaspuffar to come by and review/merge.

PR #489, opened 2021-03-06
Added QR decoder

kalaspuffar:
Issue #481.

Description:
This PR will try to solve the issue with an incorrect reading of QR Codes. It did not read the QRCode correctly because the QRCode could be encoded in multiple different ways to compress as well as possible.

* Numeric
* Alpha Numeric
* Byte stream
* Kanji
* Hanzi

As a URL seldom has Kanji or Hanzi data, I've not addressed these because I don't really understand the encoding.

For byte streams, it's pretty straightforward. They read them and add them to a list. They also add that data to a string value. Numeric or alphanumeric encoding was only added to the string value.

Changes:
My solution adds the byte stream list to the numeric and alphanumeric functions and handles the reading of these encoded values adding them to a byte stream of its own. This will, in turn, give us the expected output when we merge these streams in the application.

This might be a bug I want to report to the ZXing team, and depending on their response, we might remove this code in the future.

QRCodeDecodedBitStreamParser.java is a copy of https://github.com/zxing/zxing/blob/b1c85db64e0ef13e7d8a4c9de32bd94c76eea5d8/core/src/main/java/com/google/zxing/qrcode/decoder/DecodedBitStreamParser.java with some minor changes.

Best regards
Daniel

### DIFF
--- a/app/src/main/java/org/ea/sqrl/activites/MainActivity.java
+++ b/app/src/main/java/org/ea/sqrl/activites/MainActivity.java
@@ -9,6 +9,7 @@ import android.util.Log;
 import com.google.zxing.FormatException;
 import com.google.zxing.integration.android.IntentIntegrator;
 import com.google.zxing.integration.android.IntentResult;
+import org.ea.sqrl.utils.QRCodeDecodeHelper;
 
 import org.ea.sqrl.R;
 import org.ea.sqrl.activites.base.LoginBaseActivity;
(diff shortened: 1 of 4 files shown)


alexhauser (approved):
Nice job, Daniel! 

I've done a little testing and determined that the following works as expected:

- Scan an identity QR code
- Scan a login QR code
- Scan the specific QR code that was reported to be improperly decoded before

Looks good to me!

alexhauser:
Btw, you might consider changing the PR title to "Added <ins>QR</ins> decoder", which would be a little more self-explanatory.

PR #491, opened 2021-11-13
Improve GRADLE build Performance

shisheng-1:
[Process forking options](https://docs.gradle.org/current/userguide/performance.html#forking_options). Gradle will run all tests in a single forked VM by default. This can be problematic if there are a lot of tests or some very memory-hungry ones. We can fork a new test VM after a certain number of tests have run by setting `forkEvery`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,10 @@ buildscript {
 
 buildDir = 'gen_build'
 allprojects {
+    tasks.withType(Test).configureEach {
+        forkEvery = 100
+    }
+
     buildDir = 'gen_build'
     repositories {
         google()


kalaspuffar (approved):

kalaspuffar:
Hi @shisheng-1 

Thank you for the addition. We don't run that many tests yet, but this is a good addition if we add more tests in the future. I've not used this feature myself, so I can't vouch for its importance.

Perhaps @alexhauser knows more about this feature and could weigh in with his thoughts.

Best regards
Daniel

alexhauser:
I'm really not an expert when it comes to gradle build scripts, either.

In general, I'm not a big fan of premature optimizations or fixing non-existing (or not yet existing) problems.

@shisheng-1, can you please elaborate what adding this option will do for us at this point?

shisheng-1:
Hi @alexhauser  
I tested the time by `./gradlew test`, see the data here:
| with forkEvery | without forkEvery|
| -------| -------|
| [1m 49s](https://github.com/shisheng-1/secure-quick-reliable-login/actions/runs/1493543494)| [2m 47s](https://github.com/shisheng-1/secure-quick-reliable-login/actions/runs/1493533504)|

alexhauser:
Thanks for elaborating, @shisheng-1!
This actually does look like a substantial improvement, so LGTM @kalaspuffar!

PR #504, opened 2026-07-23
Migration

kalaspuffar:
Description:
This is just an upgrade of all components so we can compile and run on newer Android versions.

Changes:
• ✅ Updated all XML templates
• ✅ Changed the password calculator to use service in a little bit different way. 
• ✅ Migration branch builds clean (AGP 8.7.3, Gradle 8.9, compileSdk 35)
• ✅ jcenter removed entirely, replaced with mavenCentral where missing
• ✅ ProGuard keep rules in place for ZXing/lib-sodium-JNI/io.uhash


kalaspuffar (commented):
Approved

kalaspuffar:
Hi @alexhauser 

Long time no see, I'm not sure I will release on Google Play anymore but I did some work (using AI) to update all the versions so it would build for new Android versions. I think having an application on F-Droid that people could use is a benefit even if the race seems lost I still think SQRL is superior to WebAuthN. 

For one thing I had an open source application using WebAuthN that did not allow passwords so it was impossible to create a demo locally as I needed to use the phone to log in. Not great experience. SQRL would work.

Best regards
Daniel

alexhauser:
Hi @kalaspuffar,

thanks for your efforts on keeping this usable!
I can't test your changes at the moment, but would like to approve the PR non the less.
However, it seems that I can only comment but not approve.

Best regards,
Alex

alexhauser (approved):
Oh, there we go, approved!

